PR #259, opened 2015-01-12
Added Findcppnetlib.cmake file.

chevtche:


eile:
Where does the file come from?

At least _QUIETLY should be respected, and the messages should follow the standard convention. See for example https://github.com/Eyescale/CMake/blob/master/FindDNSSD.cmake.

Findcppnetlib.cmake
```diff
@@ -0,0 +1,103 @@
...
+#
+
+# Assume not found.
+SET(CPPNETLIB_FOUND FALSE)
+
+# PATH ________________________________________________________________________
+
+if (CPPNETLIB_PATH)
+    # Set by user...
+else (CPPNETLIB_PATH)
```
eile: if(NOT CPPNETLIB_PATH)

Findcppnetlib.cmake
```diff
@@ -0,0 +1,103 @@
...
+    find_path(CPPNETLIB_PATH include/boost/network.hpp
+	HINTS ${CPP_NETLIB_ROOT} $ENV{CPP_NETLIB_ROOT}
+        /usr/local/
+        /usr/
+    )
+endif (CPPNETLIB_PATH)
```
eile: use endif(), not endif([repeatedifcond])

Findcppnetlib.cmake
```diff
@@ -0,0 +1,103 @@
...
+        set(_cppnetlib_output 1)
+    endif()
+endif (CPPNETLIB_FIND_REQUIRED)
+
+include(FindPackageHandleStandardArgs)
+FIND_PACKAGE_HANDLE_STANDARD_ARGS(CPPNETLIB DEFAULT_MSG CPPNETLIB_LIBRARIES CPPNETLIB_INCLUDE_DIR)
+
+set(CPPNETLIB_INCLUDE_DIRS ${CPPNETLIB_INCLUDE_DIR})
+
+if (NOT CPPNETLIB_FOUND)
```
eile: AND _output

Findcppnetlib.cmake
```diff
@@ -0,0 +1,103 @@
...
+    message( ${_cppnetlib_output_type} "Could not find cpp-netlib")
+endif (NOT CPPNETLIB_FOUND)
```
eile: rm three lines - is already done in FindPackageHandleStandardArgs.

eile: This is still open.

Findcppnetlib.cmake
```diff
@@ -0,0 +1,93 @@
...
+#
+
+# Assume not found.
+SET(CPPNETLIB_FOUND FALSE)
+
+# PATH ________________________________________________________________________
+
+if (NOT CPPNETLIB_PATH)
+    find_path(CPPNETLIB_PATH include/boost/network.hpp
+	HINTS ${CPP_NETLIB_ROOT} $ENV{CPP_NETLIB_ROOT}
```
tribal-tec: Inconsistency in naming. CMake variable as CPPNETLIB, but the env variable has an underscore in between. I suggest to use the non-underscore one for all variables. Also document the env variable in the comment block above accordingly.

Findcppnetlib.cmake
```diff
@@ -0,0 +1,93 @@
...
+# CPPNETLIB_LIBRARIES - The cppnetlib libraries
+#==============================================================================
+#
+
+# Assume not found.
+SET(CPPNETLIB_FOUND FALSE)
+
+# PATH ________________________________________________________________________
+
+if (NOT CPPNETLIB_PATH)
```
tribal-tec: please remove the space between if and the braces in all places of this file.

Findcppnetlib.cmake
```diff
@@ -0,0 +1,93 @@
...
+
+# HEADERS AND DYNAMIC LIBRARIES_________________________________________________
+
+if (CPPNETLIB_PATH)
+    set (CPPNETLIB_INCLUDE_DIR ${CPPNETLIB_PATH}/include)
+    mark_as_advanced (CPPNETLIB_INCLUDE_DIR)
+
+    find_library(CPPNETLIB_LIBRARY1 NAMES cppnetlib-client-connections
+	HINTS ${CPPNETLIB_ROOT} $ENV{CPPNETLIB_ROOT}
+        PATHS ${CPPNETLIB_PATH}/lib64
```
tribal-tec: check for /lib as well?

Findcppnetlib.cmake
```diff
@@ -0,0 +1,93 @@
...
+
+# FOUND _______________________________________________________________________
+if (CPPNETLIB_FIND_REQUIRED)
+    set(_cppnetlib_output_type FATAL_ERROR)
+else()
+    set(_cppnetlib_output_type STATUS)
+endif ()
+
+include(FindPackageHandleStandardArgs)
+FIND_PACKAGE_HANDLE_STANDARD_ARGS(CPPNETLIB DEFAULT_MSG CPPNETLIB_LIBRARIES CPPNETLIB_INCLUDE_DIR)
```
tribal-tec: no need for UPPER_CASE function name. Lower-case also works and looks less aggressive :)

eile:
+1

PR #260, opened 2015-01-15
Separate unit from performance tests

eile:


CommonCPPCTest.cmake
```diff
@@ -52,7 +67,7 @@ foreach(FILE ${TEST_FILES})
   # Test link libraries
   target_link_libraries(${PROJECT_NAME}_${NAME}
                     ${${NAME}_LINK_LIBRARIES} ${TEST_LIBRARIES})
-  # Per target test command customisation with
+   # Per target test command customisation with
```
tribal-tec: space?

tribal-tec:
+1

CommonCPPCTest.cmake
eile: fixed

eile:
Merged with direct push.

PR #263, opened 2015-01-21
Add option to override CMake's ConfigVersion.cmake.in template.

halfflat:
CMake's write_basic_package_version_file() function uses a set of
files BasicConfigVersion-....cmake.in to configure the package
ConfigVersion.cmake file. These templates contain code that
compare the pointer word size between the calling find_package()
environment and that used when the package was built, in a
misguided attempt to enforce ABI compatibility.

For packages that just supply executables, this test is
simply counter-productive, particularly in a cross-compilation
environment.

This patch allows one to use one's own ConfigVersion template
file.

### DIFF
--- a/PackageConfig.cmake
+++ b/PackageConfig.cmake
@@ -16,7 +16,7 @@
 #     to remove from the list of exported libraries
 #   ${UPPER_PROJECT_NAME}_CONFIG_VERSION_IN - If set, use this template in
 #     preference to CMake's BasicConfigVersion-SameMajorVersion.cmake.in.
-#     The PACKAGE_VERSION variable is set before perforiming the instantiation.
+#     The PACKAGE_VERSION variable is set before performing the instantiation.
 #
 # Output variables
 #   ${UPPER_PROJECT_NAME}_FOUND - Was the project and all of the specified
--- a/PackageConfig.cmake
+++ b/PackageConfig.cmake
@@ -16,7 +16,7 @@
 #     to remove from the list of exported libraries
 #   ${UPPER_PROJECT_NAME}_CONFIG_VERSION_IN - If set, use this template in
 #     preference to CMake's BasicConfigVersion-SameMajorVersion.cmake.in.
-#     The PACKAGE_VERSION variable is set before performing the instantiation.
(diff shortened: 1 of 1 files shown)


PackageConfig.cmake
```diff
@@ -384,9 +387,14 @@ configure_file(
 )
 
 # create and install ProjectConfigVersion.cmake
-write_basic_package_version_file(
-  ${PROJECT_BINARY_DIR}/pkg/${PROJECT_NAME}ConfigVersion.cmake
-  VERSION ${VERSION_MAJOR}.${VERSION_MINOR} COMPATIBILITY SameMajorVersion)
+if(${UPPER_PROJECT_NAME}_CONFIG_VERSION_IN)
+  set(PACKAGE_VERSION "${VERSION_MAJOR}.${VERSION_MINOR}")
```
eile: Should this be here?

halfflat: PACKAGE_VERSION is set here so that it is available to the template file; analogous to the setting of CVF_PACKAGE_VERSION for BasicConfigVersion-XXXX.cmake.in.

PackageConfig.cmake
```diff
@@ -14,6 +14,9 @@
 #     are produced
 #   ${UPPER_PROJECT_NAME}_EXCLUDE_LIBRARIES - A list of library targets
 #     to remove from the list of exported libraries
+#   ${UPPER_PROJECT_NAME}_CONFIG_VERSION_IN - If set, use this template in
+#     preference to CMake's BasicConfigVersion-SameMajorVersion.cmake.in.
+#     The PACKAGE_VERSION variable is set before perforiming the instantiation.
```
tribal-tec: typo perforiming

PackageConfig.cmake
eile: This is confusing with CMAKE_PACKAGE_VERSION used in other places. Can this be delegated to the .in file or prefixed with CVF_ or PCV_?

halfflat: Sure. I'll change it to CVF_PACKAGE_VERSION.

PR #269, opened 2015-02-05
Don't override release tag for sub projects (ITK)

eile:


tribal-tec:
+1
When does this happen? We have tags called release?

eile:
ITK does. It got overwritten in config.bbp with master.

PR #270, opened 2015-02-06
Set ALL_CPP_TESTS entry from TEST_NAME

halfflat:
Changes to the setting of TEST_NAME disassociates the
list of tests in ALL_CPP_TESTS from the actual
executables. This sets the entry in ALL_CPP_TESTS
from TEST_NAME.


eile:
Good catch. thanks.

PR #273, opened 2015-02-09
Make addition of github user remote opt-in.

halfflat:
Only add a 'user' remote to a github repository
if the GIT_EXTERNAL_USER environment variable is
explicitly set.

This is a temporary work-around for an issue
where projects using Eyescake/CMake are triggering
user prompts for github credentials — essentially
the user remote functionality fails noisily.


eile:
Working on a fix right now - my idea is to only fetch origin.

PR #275, opened 2015-02-14
CMake 3.1 disallows getting certain properties from imported targets

biddisco:


biddisco:
oops, I pasted and committed a mistake, do not merge this

biddisco:
ok, fixed it. Should be ok now.

PR #276, opened 2015-02-18
Fix broken Makefiles on RHEL after #275

eile:


eile:
@chevtche : This should/might fix your Makefile issue.

tribal-tec:
+1

PR #279, opened 2015-02-24
BG/Q detection

timocafe:
hello,

I propose this modification for the BG/Q detection, it also work for BGAS


Common.cmake
```diff
@@ -129,6 +129,10 @@ if($ENV{TRAVIS})
   set(TRAVIS ON)
 endif()
 
+if(IS_DIRECTORY "/bgsys")
+    set(BGQ_A2 TRUE)
```
eile: set(BLUEGENE TRUE)

Common.cmake
```diff
@@ -129,6 +129,10 @@ if($ENV{TRAVIS})
   set(TRAVIS ON)
 endif()
 
+if(IS_DIRECTORY "/bgsys")
+    set(BLUEGENE TRUE)
```
timocafe: OK BGQ_A2 -> BLUEGEN

PR #294, opened 2015-04-20
Fix test and cppcheck targets

eile:


hernando:
+1, and indication of what was wrong may help. I understand that given any correct execution path on the old code, the new code should work. But it you ask me what this code is actually fixing, I can't really give an explanation, except that it has to do with dependencies not properly set.

eile:
Correct, make tests for example did not execute all cppcheck targets. For some reason the whole target setup was inverted from the usual natural order.

PR #297, opened 2015-04-23
Implement "cmake -DINSTALL_PACKAGES=1"

eile:
I haven't found a good place to document this feature - ideas?

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,13 +47,17 @@ and use it in your top-level CMakeLists.txt:
   documentation into PROJECT_BINARY_DIR/doc. Optional *doxygit* target to
   copy result to ../GITHUB_ORGANIZATION/Project-M.m/. Must be included
   after all targets.
-* [SubProject](SubProject.cmake): Using a top level simple CMake 
-  superproject, several CMake subprojects (which may depend on each other) 
-  can be build. Each subproject can be added to the superproject 
-  with an add_subdirectory( *subproject* ) directive. To use the SubProject
-  feature, the sub projects should modify their CMake scripts.
-  In the scripts CMAKE_BINARY_DIR should be changed to PROJECT_BINARY_DIR 
-  and CMAKE_SOURCE_DIR should be changed to PROJECT_SOURCE_DIR. A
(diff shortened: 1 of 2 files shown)


rdumusc:
Looks good, documentation for the feature can be integrated at the top of the file, and also add a line under:
# Respects the following variables:
# - INSTALL_PACKAGES: ....

eile:
Yes, that would be the natural place where nobody will find it. BlueBrain/Hello/README maybe?

I've got one more thing to test (that INSTALL_PACKAGES is not saved in CMakeCache) before merging.

eile:
Ping

Common.cmake
```diff
@@ -46,6 +46,9 @@ set(CMAKE_C_FLAGS_RELWITHDEBINFO "${CMAKE_C_FLAGS_RELWITHDEBINFO} -DNDEBUG")
 set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_FLAGS_RELWITHDEBINFO} -DNDEBUG")
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 set(CMAKE_INSTALL_MESSAGE LAZY) # no up-to-date messages on installation
+if("${CMAKE_SOURCE_DIR}" STREQUAL "${CMAKE_CURRENT_SOURCE_DIR}")
+  unset(INSTALL_PACKAGES CACHE) # Remove after install in SubProject.cmake
```
rdumusc: is there no way it can be done inside SubProject.cmake? These kind of cross-file operations variables are always hard to track and maintain...

rdumusc:
Yes, bluebrain/hello would be a good candidtate so that users can find the new feature! Eyescale/CMake also has a user documentation section which describes the subproject functionnality, but I doubt newcommers will read this. Also, the link in there should be changed to point to bluebrain/hello instead of bilgili/SubProjects.

But regardless of user documentation, please also update the developer documentation inside the cmake file, at least adding a line under: "Respects the following variables"...

eile:
both done.

rdumusc:
Perfect, +2

PR #300, opened 2015-04-28
Subproject support for cpplint.

ptoharia:


CpplintTargets.cmake
```diff
@@ -24,12 +24,12 @@ if(NOT CPPLINT_FOUND)
 endif(NOT CPPLINT_FOUND)
 
 if(NOT CPPLINT_FOUND)
-  add_custom_target(cpplint COMMENT "${CPPLINT_NOT_FOUND_MSG}")
-  set_target_properties(cpplint PROPERTIES EXCLUDE_FROM_ALL TRUE)
+  add_custom_target(cpplint_${PROJECT_NAME} COMMENT "${CPPLINT_NOT_FOUND_MSG}")
```
rdumusc: The convention is to use: ${PROJECT_NAME}_cpplint instead

rdumusc: Hmm, it looks more complex actually, I'm trying to understand why there are so many targets...

ptoharia: @rdumusc We tried to mimic what was already written for cppcheck

eile:
Superseeded by #303

PR #304, opened 2015-04-28
Add macro to rename libs for debug/release builds (on win32)

biddisco:
Tested OK on windows and Mac so far. Should do nothing on linux unless user tells it to.

### DIFF
--- a/CommonLibrary.cmake
+++ b/CommonLibrary.cmake
@@ -29,6 +29,11 @@
 # Builds libName and installs it. Installs the public headers to include/name.
 # Generates a NAME_INCLUDE_NAME/{BASE_NAME of NAME_INCLUDE_NAME}.h
 # including all public headers.
+#
+# Options 
+# By default on windows platforms "_D" is appended to libraries
+# if COMMON_LIBRARY_DEBUG_POSTFIX is ON then "_debug" is added on other platforms
+#
 
 include(InstallFiles)
 include(CommonQtSupport)
@@ -62,6 +67,19 @@ macro(generate_library_header NAME)
   # configure only touches file if changed, saves compilation after reconfigure
   configure_file( ${__generated_header_in} ${__generated_header} COPYONLY)
   list(APPEND PUBLIC_HEADERS ${__generated_header})
+endmacro()
+
+#-------------------------------------------------------------------------------
+# set debug postfix 
+#-------------------------------------------------------------------------------
+macro(COMMON_SET_LIB_NAME_POSTFIX)
+  if (WIN32)
+    set(CMAKE_DEBUG_POSTFIX "_D")
+  elseif(COMMON_LIBRARY_DEBUG_POSTFIX)
+    set(CMAKE_DEBUG_POSTFIX "_debug")
+  else()
+    set(CMAKE_DEBUG_POSTFIX "")
+  endif()
 endmacro()
 
 # applying CMAKE_C(XX)_FLAGS to add_library only works from parent scope, hence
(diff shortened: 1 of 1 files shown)


CommonLibrary.cmake
```diff
@@ -28,6 +28,66 @@ set(COMMON_LIBRARY_TYPE SHARED CACHE STRING
...
+    set (LIB_DEBUG_NAME "lib${libname}")
+    if (WIN32)
+      set (LIB_RELEASE_NAME "${libname}")
+      set (LIB_DEBUG_NAME "${libname}_D")
+    elseif (COMMON_LIBRARY_DEBUG_SUFFIX) 
+      set (LIB_RELEASE_NAME "${libname}")
+      set (LIB_DEBUG_NAME "${libname}_debug")
+    endif ()
+  else (${libtype} MATCHES "SHARED")
+    # default for linux/mac/unix
```
rdumusc: The defaults are the same for shared and non-shared, could be moved outside of the if() else()
General style comments:
- no space between operator and parenthesis: if() not if ()
- no repetition of the condition in the else() and endif() (keep them empty)

CommonLibrary.cmake
```diff
@@ -28,6 +28,66 @@ set(COMMON_LIBRARY_TYPE SHARED CACHE STRING
...
+  # message (STATUS "${libname} libtype: ${libtype}")
+  if (${libtype} MATCHES "SHARED")
+    # default for linux/mac/unix
+    set (LIB_RELEASE_NAME "lib${libname}")
+    set (LIB_DEBUG_NAME "lib${libname}")
+    if (WIN32)
+      set (LIB_RELEASE_NAME "${libname}")
+      set (LIB_DEBUG_NAME "${libname}_D")
+    elseif (COMMON_LIBRARY_DEBUG_SUFFIX) 
+      set (LIB_RELEASE_NAME "${libname}")
```
rdumusc: Unless I'm mistaken, this would generate unix libs without the lib prefix, is this intended?

CommonLibrary.cmake
```diff
@@ -28,6 +28,66 @@ set(COMMON_LIBRARY_TYPE SHARED CACHE STRING
...
+    # default for linux/mac/unix
+    set (LIB_RELEASE_NAME "lib${libname}")
+    set (LIB_DEBUG_NAME "lib${libname}")
+    if (WIN32)
+      set (LIB_RELEASE_NAME "lib${libname}")
+      set (LIB_DEBUG_NAME "lib${libname}_D")
+    elseif (COMMON_LIBRARY_DEBUG_SUFFIX)
+      # if the generator supports configuration types or if the CMAKE_BUILD_TYPE has a value
+      if (CMAKE_CONFIGURATION_TYPES OR CMAKE_BUILD_TYPE)
+        set (LIB_RELEASE_NAME "${libname}")
```
rdumusc: Same question here, what is the use case for this distinction with or without lib prefix on non-win32 platforms?

rdumusc:
Last thing: add a line at line 16 documenting the use of COMMON_LIBRARY_DEBUG_SUFFIX.
Good otherwise!

rdumusc:
@biddisco waiting for your updated version, please push when you have time :-)

eile:
@rdumusc biddisco/CMake#1 is the updated version. Waiting for John to review.

biddisco:
@rdumusc made a new version that is very simple. Rebased it onto eile:master not much point me doing PRs as they all get chucked out anyway

biddisco:
rebased onto master

PR #305, opened 2015-04-29
Use correct property to exclude targets from VS "All Build"

eile:
Also unified use of ON vs TRUE.


rdumusc:
+1! Good for the unification even though I preferred TRUE ;-))

PR #307, opened 2015-05-01
target warnings in ninja, explicit config superprojects

eile:
Re 931bf29: The always-automatic include of *cmake is dangerous for projects which do have cmake files in the top-level directory (hello @biddisco). config superprojects have to explicitly call subproject_configure (see https://bbpcode.epfl.ch/code/#/c/11763/).


tribal-tec:
+2

PR #308, opened 2015-05-04
Restore missing cppcheck include for version.h (remove in #284)

rdumusc:
@tribal-tec FYI - This folder was removed from includes in #284 . This breaks the compilation of BBPSDK after the update:

https://bbpcode.epfl.ch/ci/job/common.BBPSDK.gerrit/build_type=Debug,platform=cscsviz/261/testReport/projectroot.gpfs.bbpcscsch.home.bbprelman.jenkins-cscs-02.workspace.commonBBPSDKgerrit.build_type.Debug.platform.cscsviz/BBP/cppcheck_test_BBPSDK/

### DIFF
--- a/TargetHooks.cmake
+++ b/TargetHooks.cmake
@@ -11,8 +11,10 @@ include(CpplintTargets)
 
 set(ALL_DEP_TARGETS "")
 set(ALL_LIB_TARGETS "")
+
+# ${PROJECT_NAMESPACE}_API= -> Fix cppcheck error about not including version.h
 set(CPPCHECK_EXTRA_ARGS
-  -D${UPPER_PROJECT_NAME}_STATIC= -D${UPPER_PROJECT_NAME}_API=)
+  -D${UPPER_PROJECT_NAME}_STATIC= -D${PROJECT_NAMESPACE}_API=)
 
 # only ever define this macro once, in case subprojects include the same rules
 get_property(ADD_EXE_DEFINED GLOBAL PROPERTY ADD_EXE_MACRO_DEFINED)


tribal-tec:
I don't see the error. Besides, iirc, this was one of the main speedups (especially noticeable in Equalizer) for the cppcheck run. If you can ensure it is not 10x slower, we can merge this, otherwise i strongly disagree.

rdumusc:
I haven't checked the difference in execution time, but could you please detail what you mean when you say you are not seeing the error? cppcheck now fails with the following output:

/gpfs/bbp.cscs.ch/home/bbprelman/jenkins-cscs-02/workspace/common.BBPSDK.gerrit/build_type/Debug/platform/cscsviz/build/include/BBP/version.h:56: information: The code 'class BBP_API Version {' is not handled. You can use -I or --include to add handling of this code. 

It is something I can replicate locally... do you have any idea how else I could fix that?

tribal-tec:
In Equalizer I undef'd the API: list(APPEND CPPCHECK_EXTRA_ARGS -DEQUALIZER_DSO_NAME=foo -DEQ_API=)

rdumusc:
Thanks, this helps! It is what TargetHooks attempts to do then:
set(CPPCHECK_EXTRA_ARGS -D${UPPER_PROJECT_NAME}_STATIC= -D${UPPER_PROJECT_NAME}_API=)
except that it doesn't work for BBPSDK because UPPER_PROJECT_NAME (BBPSDK) != PROJECT_INCLUDE_NAME (BBP). I will submit a new patchset to fix this.

rdumusc:
Updated

tribal-tec:
+1

PR #309, opened 2015-05-06
Fix issues with 'tests' target

eile:
* d81529d by @hernando started support for multiple test suites. This
  caused higher-level cppcheck targets not to run anymore, and in
  reality only worked for one instance of cpp ctests.
* Each test is now run as a separate ctest (side benefit: parallel tests
  as per '-j' build command line!). This also simplifies perf/unit test
  separation
* Directly add run_cppcheck targets to tests target instead of using
  add_test.
* Doxygen depends only on project tests for coverage, not on all tests
* Remove cppcheck and cpplint tests, covered by run_ targets now.
* Jenkins will need to change to use 'make tests' instead of ctest
  directly. ctest command line parameters needed for this change are
  included here


eile:
@hernando @rdumusc @tribal-tec FYI/please review

tribal-tec:
Reviewed offline already. +1.

CppcheckTargets.cmake
```diff
@@ -125,6 +117,11 @@ function(add_cppcheck _name)
   if(NOT TARGET cppcheck_${PROJECT_NAME})
     add_custom_target(cppcheck_${PROJECT_NAME})
   endif()
+  if(NOT TARGET ${PROJECT_NAME}_tests)
```
rdumusc: target "${PROJECT_NAME}_tests" can now be added both in here and in CommonCTest.cmake, but only the latter does "set_target_properties()". Can it be added only in one place?

CppcheckTargets.cmake
```diff
@@ -125,6 +117,11 @@ function(add_cppcheck _name)
...
+    add_custom_target(${PROJECT_NAME}_tests)
+  endif()
+
   add_dependencies(cppcheck_${PROJECT_NAME} cppcheck_run_${_name})
+  add_dependencies(${PROJECT_NAME}_tests cppcheck_run_${_name})
```
rdumusc: Is there a good reason for having cppcheck_run_${_name} part of two meta-targets..?

CppcheckTargets.cmake
rdumusc: and CpplintTargets.cmake does not do it. I would remove it from here, or else it must also be added in the CpplintTargets.cmake...

CommonCPPCTest.cmake
```diff
@@ -82,11 +82,17 @@ macro(common_add_cpp_test NAME FILE)
   add_test(NAME ${TEST_NAME}
     COMMAND ${RUN_PREFIX} $<TARGET_FILE:${TEST_NAME}> ${RUN_ARGS})
 
+  add_custom_target(ctest_${TEST_NAME}
```
rdumusc: are set_target_properties(  EXCLUDE_FROM_DEFAULT_BUILD ON FOLDER "Tests" ) not needed for these targets ?

CppcheckTargets.cmake
eile: It needs to be added here and in cpplint.

I have not found a better way :(

CppcheckTargets.cmake
eile: To be able to do 'ninja cppcheck' and 'ninja tests'.

CommonCPPCTest.cmake
```diff
@@ -50,9 +50,9 @@ macro(common_add_cpp_test NAME FILE)
   endif()
 
   if(NAME MATCHES "^perf.*")
-    list(APPEND ALL_CPP_PERF_TESTS ${TEST_NAME})
+    list(APPEND ALL_CPP_PERF_TESTS ctest_${TEST_NAME})
   else()
-    list(APPEND ALL_CPP_TESTS ${TEST_NAME})
+    list(APPEND ALL_CPP_TESTS ctest_${TEST_NAME})
```
rdumusc: ctest_${TEST_NAME} target is appended to the list before being defined, which is a bit confusing. I would suggest moving it after add_custom_target(ctest_${TEST_NAME}).

CommonCPPCTest.cmake
eile: done

CppcheckTargets.cmake
eile: done for cpplint

CommonCPPCTest.cmake
eile: done

hernando:
Is the documentation at the top of the modified scripts up to date?

hernando:
This change breaks RTNeuron because of the approach used to have both cpp and Python tests. I think it's better to discuss it offline.

hernando:
I fixed some issues in RTNeuron but coverage_report cannot be called multiple types, CMake give an error about repeated target names. This function needs to be called multiple times to simplify the addition of multiple test suites.

eile:
Documentation is up to date

eile:
RTNeuron: Skype?

CppcheckTargets.cmake
rdumusc: Why not let tests depend on cppcheck to achieve this?

CppcheckTargets.cmake
rdumusc: OK, this is very sad... :-(

rdumusc:
I think documentation could be improved by adding an exhaustive list of the targets created by each file (bullet point style) and their relation to targets created by other files. I can try to assist with this if you want.

eile:
Ping @hernando please test with RTNeuron and +/-1 accordingly.

hernando:
+1

PR #321, opened 2015-05-26
Use Module mode for find_package in subprojects if Config mode fails.

hernando:
This is needed to correctly detect subprojects that can be system level installed
but provide no <project>Config.cmake files (e.g. cppnetlib).


tribal-tec:
+1

PR #322, opened 2015-05-26
Fix doxygen_install, cleanup Coverage target and force LCOV 1.11

tribal-tec:
FYI @eile

### DIFF
--- a/DoxygenRule.cmake
+++ b/DoxygenRule.cmake
@@ -85,8 +85,8 @@ if(NOT DOXYGEN_CONFIG_FILE)
 endif()
 
 add_custom_target(doxygen_install_${PROJECT_NAME}
-  ${CMAKE_COMMAND} -DCMAKE_INSTALL_LOCAL_ONLY=ON
-  -P ${PROJECT_BINARY_DIR}/cmake_install.cmake
+  ${CMAKE_COMMAND}
+  -P ${PROJECT_BINARY_DIR}/${PROJECT_INCLUDE_NAME}/cmake_install.cmake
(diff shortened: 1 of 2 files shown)


rdumusc:
+1, tested with both Buildyard and Subproject.
Lcov 1.11 is needed because 1.10 issues an error instead of a warning for the --directory PROJECY_SOURCE_DIR, which (obviously) contains no .gcda files.

DoxygenRule.cmake
```diff
@@ -85,8 +85,8 @@ if(NOT DOXYGEN_CONFIG_FILE)
 endif()
 
 add_custom_target(doxygen_install_${PROJECT_NAME}
-  ${CMAKE_COMMAND} -DCMAKE_INSTALL_LOCAL_ONLY=ON
-  -P ${PROJECT_BINARY_DIR}/cmake_install.cmake
+  ${CMAKE_COMMAND}
+  -P ${PROJECT_BINARY_DIR}/${PROJECT_INCLUDE_NAME}/cmake_install.cmake
```
eile: Why? This installs only the project headers, but not for example examples.

rdumusc: Examples are not needed for generating documentation, are they? Only the headers are, so the minimal subset is installed which is also faster. It is not possible to install everything anyway because then it tries to also install binaries of subrojects, which have not been build (only subproject libs are built if you do make doxygen_PROJECTNAME, so failure).

eile: At least in Equalizer the examples are also doxygen-ed. INSTALL_DEPENDS should include applications imo. That used to work at least.

PR #323, opened 2015-05-27
More QUIET for internal find_packages, remove obsolete GNUModules

tribal-tec:


CommonCPack.cmake
```diff
@@ -45,7 +45,6 @@ set(CPACK_PACKAGE_VERSION ${VERSION})
 set(CPACK_RESOURCE_FILE_LICENSE ${PROJECT_SOURCE_DIR}/LICENSE.txt)
 if(NOT CPACK_PACKAGE_LICENSE)
   set(CPACK_PACKAGE_LICENSE "LGPL")
-  message(STATUS "Using default ${CPACK_PACKAGE_LICENSE} package license")
```
eile: Is this a good idea? I would rather make it a FATAL_ERROR, because declaring a license for a project automatically seems wrong.

eile:
+1 except license.

Add option(COMMON_PACKAGE_QUIET ON)

tribal-tec:
Updated

PR #325, opened 2015-05-29
Fixes in Python autodetection

hernando:


tribal-tec:
Just have one if(TARGET coverage_${PROJECTNAME}) and then early-out? The four targets are added once or not at all imo.

rdumusc:
I'm not sure, a single test could work but probably without the early out because the idea is to do the add_dependencies(${TEST_TARGET}) after making sure that the targets have been created...

tribal-tec:
True. It's just not helping to read it with the four if statements and indentation.

hernando:
I also thought of doing a single test, but since this scheme is used in many other parts I decided to follow the "style".

PR #326, opened 2015-06-01
Fixed the Findcppnetlib script

bilgili:
The cppnetlib finder was broken. It could either find includes or libraries.


rdumusc:
+1

PR #329, opened 2015-06-05
Added common_bundle() function for OSX

rdumusc:


eile:
Why not merge bundle into application, e.g., as an option (or even per default?)

rdumusc:
Actually not a bad idea! Not by default (generally common_application() should do a regular command line application) but we could apply the extra steps if MACOSX_BUNDLE was passed in ARGV. Don't merge yet, I will prepare a new patch for that.

eile:
That's exactly the question: Is command line or UI application the "default"? Shall we make it dependent on the presence of Qt?

rdumusc:
updated

CommonApplication.cmake
```diff
@@ -13,6 +13,13 @@
 # ** NAME_UI_FORMS list of all .ui input files
 # ** NAME_RESOURCES list of all .qrc resource files
 #
+# Cross-plateform wrapper around common_application() for GUI applications:
```
eile: Frypo: Platform

eile:
+1

CommonApplication.cmake
rdumusc: Thanks, I keep doing this frypo :-)

PR #338, opened 2015-07-03
Fixing Python autodetection and adding a sanity check.

hernando:
The script fails when Boost.Python3 is detected but Python 3 is not really
available.
The error is a regression from e93d8438.
PYTHON_EXECUTABLE mustn't be used because the PythonInterp script sets the
variable to any interpreter found even it has failed because the version
doesn't match the requested one.


hernando:
FYI @tribal-tec

PR #339, opened 2015-07-08
common_gui_application() links to Qt5::WinMain for Qt apps on windows

rdumusc:
I'm actually not sure about the elseif(WIN32) or elseif(MSVC). Is the first one not valid? If that's the case, then the wrong code path was taken and WIN32 was not even added to the executable, and there is a change that simply changing that line could be sufficient to solve the problem, without linking to WinMain. Could you please check that too?

### DIFF
--- a/CommonApplication.cmake
+++ b/CommonApplication.cmake
@@ -40,9 +40,9 @@ function(COMMON_APPLICATION Name)
 endfunction()
 
 function(COMMON_GUI_APPLICATION Name)
-if(APPLE)
-  string(TOUPPER ${Name} NAME)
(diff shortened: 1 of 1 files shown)


CommonApplication.cmake
```diff
@@ -70,6 +70,11 @@ if(APPLE)
     WORKING_DIRECTORY ${_INSTALLDIR})" COMPONENT apps)
 elseif(WIN32)
```
tribal-tec: MSVC. WIN32 means also cygwin, where this is probably not needed as it behaves like Linux.

rdumusc: Thanks! updated

PR #342, opened 2015-07-13
Added a cached variable to choose where to download subproject sources.

hernando:


SubProject.cmake
```diff
@@ -48,6 +50,12 @@ add_custom_target(git_subproject_${PROJECT_NAME}_done)
 set_target_properties(git_subproject_${PROJECT_NAME}_done PROPERTIES
   EXCLUDE_FROM_DEFAULT_BUILD ON FOLDER "zzphony")
 
+set(ROOT_SOURCE_DIR "${CMAKE_SOURCE_DIR}" CACHE PATH
+  "Location of root directory of all source")
```
tribal-tec: all sources

SubProject.cmake
```diff
@@ -31,6 +31,8 @@
 #   example for continuous integration builds
 # - INSTALL_PACKAGES: command line cache variable which will "apt-get" or
 #   "port install" the known system packages. Will be unset after installation.
+# - ROOT_SOURCE_DIR: When set, the source code of subprojects will be
+#   downloaded in this path instead of CMAKE_SOURCE_DIR.
```
eile: Please call it COMMON_SOURCE_DIR (consistency with other common stuff, plus use case is inherently clear)

hernando:
Will try to merge with #342.

PR #348, opened 2015-07-15
Make external update an explicit target rather then an implicit cmake step

eile:


eile:
@rdumusc FYI

GitExternal.cmake
```diff
@@ -101,67 +109,68 @@ function(GIT_EXTERNAL DIR REPO TAG)
...
+    "    ERROR_QUIET OUTPUT_QUIET\n"
+    "    WORKING_DIRECTORY \"${DIR}\")\n"
+    "  execute_process(\n"
+    "    COMMAND \"${GIT_EXECUTABLE}\" checkout -q -- \"${GIT_EXTERNAL_RESET_FILE}\"\n"
+    "    ERROR_QUIET OUTPUT_QUIET\n"
+    "    WORKING_DIRECTORY \"${DIR}\")\n"
+    "endforeach()\n"
+    "\n"
+    # fetch latest update
+    "execute_process(COMMAND \"${GIT_EXECUTABLE}\" fetch origin -q\n"
```
tribal-tec: make 'origin' a configurable variable? I'd like to fetch from 'user' or even '-all' sometimes.

GitExternal.cmake
```diff
@@ -101,67 +109,68 @@ function(GIT_EXTERNAL DIR REPO TAG)
...
+    "endif()\n"
+    "\n"
+    # update tag
+    "execute_process(COMMAND ${GIT_EXECUTABLE} rebase FETCH_HEAD\n"
+    "  RESULT_VARIABLE nok ERROR_VARIABLE error OUTPUT_VARIABLE output\n"
+    "  WORKING_DIRECTORY \"${DIR}\")\n"
+    "if(nok)\n"
+    "  execute_process(COMMAND ${GIT_EXECUTABLE} rebase --abort\n"
+    "    WORKING_DIRECTORY \"${DIR}\" ERROR_QUIET OUTPUT_QUIET)\n"
+    "  message(FATAL_ERROR \"Rebase ${__dir} failed:\n\${output}\${error}\")\n"
```
tribal-tec: does it need to be FATAL_ERROR here?

GitExternal.cmake
eile: New feature -> in a new commit/PR.

GitExternal.cmake
eile: Now that it's an explicit target I'ld like it to fail on failure. One can always do 'ninja rebase -k0'

tribal-tec:
Ok, +1 for me. I suggest you wait for @rdumusc though :)

GitExternal.cmake
```diff
@@ -14,13 +14,12 @@
 #    or in the given absolute path using the given repository and tag
 #    (commit-ish).
 #
+# Targets:
+#  * DIR-rebase: fetches latest updates and rebases the given external
```
rdumusc: rename DIR-rebase to <directory>-rebase for consistency with the rest of the documentation

GitExternal.cmake
```diff
@@ -101,67 +109,68 @@ function(GIT_EXTERNAL DIR REPO TAG)
...
+    )
 
-  # checkout requested tag
-  execute_process(
-    COMMAND "${GIT_EXECUTABLE}" checkout -q "${TAG}"
-    RESULT_VARIABLE nok ERROR_VARIABLE error
-    WORKING_DIRECTORY "${DIR}")
-  if(nok)
-    message(STATUS "git checkout ${TAG} in ${DIR} failed: ${error}\n")
+  add_custom_target(${__target}-rebase
```
rdumusc: Make these targets Windows friendly: EXCLUDE_FROM_DEFAULT_BUILD ON FOLDER "git" or something like that

eile:
updated

rdumusc:
Other than that, looks good to me! I would also propose to have all git-related target share a common prefix, such as git_rebase_all, git_update_externals, etc... but I have the feeling you won't like that idea ;-)

eile:
definitely not with '_'. :P

Maybe in a later commit. Targets are still a mess, let's fix this one by one.

PR #352, opened 2015-07-15
Fixing compilation flags on Blue Gene/Q and adding better level of optimization

delalond:
Updating compilation flags for XLC compiler on Blue Gene/Q. Higher level of optimization added. Removed dynamic linking by default which was causing issues for some software. Dynamic linking can still be added on a case by case basis by individual project.


eile:
Needs rebase to merge. +1 otherwise.

eile:
Can be closed?

PR #358, opened 2015-07-16
Only add user forks if GITHUB_USER (cmake or env) variable is set

eile:


eile:
@tribal-tec @biddisco FYI

GitExternal.cmake
```diff
@@ -100,13 +93,12 @@ function(GIT_EXTERNAL DIR REPO TAG)
   endif()
 
   # set up "user" remote for github forks
-  if(GIT_EXTERNAL_USER_FORK AND REPO MATCHES ".*github.com.*")
-    string(REGEX REPLACE "(.*github.com[\\/:]).*(\\/.*)"
-      "\\1${GIT_EXTERNAL_USER_FORK}\\2" GIT_EXTERNAL_USER_REPO ${REPO})
+  if(${GITHUB_USER AND REPO MATCHES ".*github.com.*")
```
rdumusc: missing closing }, but it think it should it be without the ${}

rdumusc:
Very nice, +1 once the above is fixed!

GitExternal.cmake
eile: Nice catch.

PR #363, opened 2015-07-16
Fix GITHUB_USER fork using ENV variable

rdumusc:


rdumusc:
Shouldn't we also change the remote to be ssh instead of httpss? This one is kind of unusable because you have to type your credentials all the time...

tribal-tec:
If your behind a firewall, you can't ssh. https will always work. For pull, you don't need ssh, but agreed, for push I also want to use ssh. It should be a configurable option which protocol should be used.

eile:
```
[skywriter common master]> cat ~/.gitconfig
[user]
	name = Stefan Eilemann
	email = Stefan.Eilemann@epfl.ch
[color]
	ui = auto
[credential]
	helper = cache --timeout=36000
```

PR #370, opened 2015-07-20
Prepend compiler flags to let users disable compiler warnings

rdumusc:


tribal-tec:
Interesting. We didn't do it like this before imo. You have an example where this applies?

rdumusc:
Yes I do :-) look how https://github.com/BlueBrain/TUIO/blob/master/CMakeLists.txt#L35 was meant to override https://github.com/rdumusc/CMake/blob/master/Compiler.cmake#L81

Without this PR, on OS X you get:
[ 28%] Building CXX object TUIO/CMakeFiles/SimpleSimulator.dir/SimpleSimulator.cpp.o
In file included from /Users/dumusc/src/DisplayCluster/TUIO/SimpleSimulator.cpp:22:
/Users/dumusc/src/DisplayCluster/TUIO/SimpleSimulator.h:43:17: error: using namespace directive in
      global context in header [-Werror,-Wheader-hygiene]
using namespace TUIO;
                ^
/Users/dumusc/src/DisplayCluster/TUIO/SimpleSimulator.h:74:6: error: private field 's_id' is not used
      [-Werror,-Wunused-private-field]
        int s_id;
            ^
2 errors generated.

PR #371, opened 2015-07-21
rm DeadCode; re-add default build packages

eile:


tribal-tec:
Update CHANGES.md? I now also means, > 2.8.8 is required.
Also, OSSCPack is still used imo (Collage, Equalizer, ...)

eile:
All done.

Wasn't >2.8.8 required for something else already?

tribal-tec:
Don't know. Most of the projects demand only 2.8, so it's hard to tell.

eile:
@rdumusc FYI

tribal-tec:
+1

rdumusc:
+1, in the future we might also add doxygen and graphviz to  ${NAME}_PORT_DEPENDS - although it is not strictly needed for building, you needed it for the documentation targets

rdumusc:
oh, I see that CMakePackageConfigHelpers is still included here: https://github.com/Eyescale/CMake/blob/master/PackageConfig.cmake#L50, should be removed, right?

and here, even though flatbuffers doesn't have a subtree!? : https://github.com/BlueBrain/flatbuffers/blob/master/CMakeLists.txt#L167

rdumusc:
how does this even work? I don't see CMakePackageConfigHelpers anywhere in my CMAKE_PATH, yet cmake does not complain?

rdumusc:
sorry, got it - its built-in from cmake...

PR #372, opened 2015-07-21
Fix SubProject find_package to work with /cmake/lib/project paths

biddisco:
Some projects put their ${project}Config.cmake files in locations
other than the root. The recent changes to SubProjects added an
automatic find and do not look in those locations.

### DIFF
--- a/SubProject.cmake
+++ b/SubProject.cmake
@@ -126,7 +126,11 @@ function(add_subproject name)
     add_subdirectory("${__common_source_dir}/${path}"
                      "${CMAKE_BINARY_DIR}/${name}")
     if(NOT ${name}_SKIP_FIND)
-      find_package(${name} REQUIRED QUIET) # find subproject "package"
+      # find subproject "package"       
+      find_package(${name} REQUIRED QUIET
+        HINTS "${CMAKE_BINARY_DIR}/${name}"
+              "${CMAKE_BINARY_DIR}/${name}/lib/cmake/${name}"
+      )
       include_directories(${${NAME}_INCLUDE_DIRS})
     endif()
     set(${name}_IS_SUBPROJECT ON PARENT_SCOPE)


eile:
John, I will have to revert this. It breaks the Livre build since Tuvok is not found anymore with this change (FindTuvok is not read at all). Can you try to find another way, e.g., setting <package>_DIR?

eile:
Which project is this? How can I reproduce?

hernando:
@eile I was having the same problem today, BlueBrain/Tuvok#7 fixes the problem (and render FindTuvok unnecesary)

rdumusc:
@hernando and BlueBrain/Tuvok#7 works because CommonCPack it indirectly includes PackageConfig, which is now required for Subproject compatibility (although not documented anywhere!). I think that FindTuvok is still needed to locate the module on gpfs, unless it is rebuilt with the PackageConfig included.
I have no idea why adding the HINTS above broke the FindTuvok though...

biddisco:
Adding the hints should only increase the number of places searched, so it should not break anything.
As it turns out, the project that was failing was HPX and I spent some time yesterday trying to fix other build issues with it and discovered that there was a case hpx or HPX mismatch in the build setup which is why it failed originally, so in fact my patch might not bee needed after all anyway. If you want to revert it, go ahead and I'll get stuff working another way.

eile:
Well, for some reasons the HINTS did what they shouldn't do. In the cmake --trace I saw 'find_package(Tuvok)' which did not jump into FindTuvok.cmake, even though it was in the CMAKE_MODULE_PATH

biddisco:
Revert it then.
One thing though. When you add_subproject, you are adding it. Calling find_package inside add_subproject - for the project you are adding is really just wrong. I understand that it is a convenience so that you maybe don't have to call it elsewhere, but it is wrong and it will cause trouble later. find_package should be called by packages that need the subproject, not by the subproject itself.

PR #373, opened 2015-07-22
Fix find_package using <prefix>/cmake/lib/<project> config path

biddisco:
```
In the cmake documentation
http://www.cmake.org/cmake/help/v3.2/manual/cmake-packages.7.html
it states that "One place it looks is: <prefix>/lib/cmake/Foo*/ where
Foo* is a case-insensitive globbing expression?

This solves the problem of projects that generate <project>Config.cmake in
the root, or in the lib/cmake/<project> location at build time.
```


eile:
Is this still relevant?

rdumusc:
This has been inactive for 2 years, closing.

PR #379, opened 2015-07-27
Libray and include paths are defined for upper and lower cases packag…

chevtche:
…e name.


rdumusc:
+1 for me, but can we also remove ${Package_Name}_name? From what I can see, it is not used anywhere outside this file, right?

rdumusc:
Oh, looks like it is needed by PackageConfig so forget my previous comment - sadly this is another case of a fully undocumented requirement...

tribal-tec:
+1

CommonPackage.cmake
```diff
@@ -132,20 +132,25 @@ macro(common_package Package_Name)
...
     set(${PROJECT_NAME}_FIND_PACKAGES_FOUND
       "${${PROJECT_NAME}_FIND_PACKAGES_FOUND} ${Package_Name}")
-    link_directories(${${${Package_Name}_name}_LIBRARY_DIRS})
-    if(NOT "${${${Package_Name}_name}_INCLUDE_DIRS}" MATCHES "-NOTFOUND")
-      if(__is_system_package)
-        include_directories(BEFORE SYSTEM ${${${Package_Name}_name}_INCLUDE_DIRS})
-      else()
-        include_directories(${${${Package_Name}_name}_INCLUDE_DIRS})
-      endif()
+    set(__library_dirs ${Package_Name}_LIBRARY_DIRS ${PACKAGE_NAME}_LIBRARY_DIRS)
```
eile: Why not set(${PACKAGE_NAME}_LIBRARY_DIRS ${PACKAGE_NAME}_LIBRARY_DIRS ${Package_Name}_LIBRARY_DIRS)?

CommonPackage.cmake
```diff
@@ -132,20 +132,25 @@ macro(common_package Package_Name)
...
-      endif()
+    if(NOT "${${PACKAGE_NAME}_INCLUDE_DIRS}" MATCHES "-NOTFOUND")
+      list(APPEND __include_dirs ${${PACKAGE_NAME}_INCLUDE_DIRS})
+    endif()
+    if(NOT "${${Package_Name}_INCLUDE_DIR}" MATCHES "-NOTFOUND")
+      list(APPEND __include_dirs ${${Package_Name}_INCLUDE_DIR}) 
+    endif()
+    if(NOT "${${PACKAGE_NAME}_INCLUDE_DIR}" MATCHES "-NOTFOUND")
+      list(APPEND __include_dirs ${${PACKAGE_NAME}_INCLUDE_DIR}) 
+    endif()
```
eile: Why not set(${${PACKAGE_NAME}_INCLUDE_DIRS} ${${PACKAGE_NAME}_INCLUDE_DIRS} + three other variants)?

CommonPackage.cmake
rdumusc: I thought about it too, but this has potential side-effects, and what benefit would there be to affect the global variables? The folders are all added in here, what more can be done outside of here?

eile: The benefit is that client code can always use NAME_FOO

eile: IC, but then we'll need to remove the _directories() from the outside cmake code.

rdumusc: But is there other client code for these two variables? they are only consumed here, right?

PR #380, opened 2015-07-27
Fix common_package() cache bug wrt NAME_FOUND

tribal-tec:
In the chaos of find_package and common_package, called by SubProject,
PackageConfig and each project, it cannot be reliably assumed that
NAME_FOUND guarantees to have NAME_LIBRARIES and other variables be
still set. Hence a find_package is always performed while calling
common_package() to avoid remedy sideffects.


tribal-tec:
Updated. Reset the local variables now.

eile:
+1. Hard to review.

PR #381, opened 2015-07-28
Restore Findcppnetlib to fix Ubuntu 14.04 build (missing CMake config…

rdumusc:
… in system package).

Does not break OSX build.

### DIFF
--- a/Findcppnetlib.cmake
+++ b/Findcppnetlib.cmake
@@ -1,0 +1,102 @@
+#
+# Copyright 2015 Grigori Chevtchenko <grigori.chevtchenko@epfl.ch>
+#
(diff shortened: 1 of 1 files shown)


tribal-tec:
-2. Cppnetlib provides cppnetlibConfig.cmake. This finder clashes with this. Ubuntu package of libcppnetlib-dev has no package config. It is fixed in 15.04. See http://packages.ubuntu.com/vivid/amd64/libcppnetlib-dev/filelist

rdumusc:
Why does it clash? cppnetlibConfig.cmake should take precedence if it is present, but otherwise I think that this is valid fallback. How do we support Ubuntu 14.04 otherwise?

eile:
Merge?

PR #382, opened 2015-07-28
GitExternals: Always print git TAG when cloning

rdumusc:
Useful to see which version of the CMake/common is used (by Jenkins for instance)

### DIFF
--- a/GitExternal.cmake
+++ b/GitExternal.cmake
@@ -95,9 +95,11 @@ function(GIT_EXTERNAL DIR REPO TAG)
     set(_clone_options --recursive)
     if(GIT_EXTERNAL_LOCAL_SHALLOW)
       list(APPEND _clone_options --depth 1 --branch ${TAG})
+    else()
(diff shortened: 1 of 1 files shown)


GitExternal.cmake
```diff
@@ -95,9 +95,11 @@ function(GIT_EXTERNAL DIR REPO TAG)
     set(_clone_options --recursive)
     if(GIT_EXTERNAL_LOCAL_SHALLOW)
       list(APPEND _clone_options --depth 1 --branch ${TAG})
+    else()
+      set(_msg_tag "[${TAG}]")
     endif()
     JOIN("${_clone_options}" " " _msg_text)
-    message(STATUS "git clone ${_msg_text} ${REPO} ${DIR}")
+    message(STATUS "git clone ${_msg_text} ${REPO} ${_msg_tag} ${DIR}")
```
eile: Move after dir? That way the print matches the command line up to []

rdumusc: OK, updated

PR #385, opened 2015-07-29
GitExternal.cmake: 'user' remote default for push & use ssh URL

tribal-tec:


rdumusc:
-1 for the asymmetry between push and pull. I'm worried it is encouraging users to do 'git push' without thinking, just to save typing a few letters. And when you clone a project directly, you don't have the 'user' remote set for the main project and the push ends up on 'origin'.
+1 for the rest :-)

tribal-tec:
Nobody forces you to use the user fork. The asymmetry always happen with Github + fork workflow (https://github.com/blog/2042-git-2-5-including-multiple-worktrees-and-triangular-workflows), so now it's just reflecting what your are supposed to do.

eile:
+1

PR #391, opened 2015-08-05
Require cppcheck version 1.66 which has better support for C++11

tribal-tec:


tribal-tec:
FYI @rdumusc

rdumusc:
+1, upgrade already required by C++11 usage in ZeroBuf, and consistent with cluster module. As a drawback Ubuntu users will be forced to compile their own cppcheck as cppcheck >= 1.68 won't happen before 15.10 (wily) is released: http://packages.ubuntu.com/search?keywords=cppcheck

rdumusc:
Or should we lower requirements to 1.66?

tribal-tec:
There are two false-positives in Collage with 1.66. I can 'fix' it with an inline suppress comment. Then we can do 1.66.

tribal-tec:
Updated

rdumusc:
+1 it will be much more friendly for the many Ubuntu users out there

PR #392, opened 2015-08-05
Add check targets (clangcheck, cppcheck, cpplint) in CommonApplication/Library

tribal-tec:


tribal-tec:
FYI @rdumusc

rdumusc:
+1 looks correct and will avoid some troubles with external projects

eile:
+10

PR #393, opened 2015-08-05
Warn users of TestCPP11 if it runs multiple times (for each subproject)

rdumusc:
It increases verbosity by one line, but it allows to spot inefficient multiple inclusions. In DisplayCluster for instance, Subproject was included before Common (right after GitExternals), and as a result TESTS_CPP11 was run for every subproject, then again for the superproject

### DIFF
--- a/TestCPP11.cmake
+++ b/TestCPP11.cmake
@@ -7,6 +7,11 @@ if(TESTS_CPP11_DONE)
 endif()
 set(TESTS_CPP11_DONE ON)
 
+if(NOT PROJECT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
+  message(WARNING "For better performance, include TestCPP11 in the top-level "
+    "project so that it is run only once.")
+endif()
+
 set(TESTS_CPP11 sharedptr tuple auto nullptr array final_override unordered_map
   template_alias)
 


TestCPP11.cmake
```diff
@@ -98,6 +100,13 @@ while(TESTS_CPP11)
...
+    list(APPEND __supported ${TEST_CPP11_name})
+  else()
+    list(APPEND __unsupported ${TEST_CPP11_name})
   endif()
 endwhile()
 
+if(__unsupported)
+  set(__unsupported " UNSUPPORTED: ${__unsupported}")
+endif()
+message(STATUS "Tested C++11 features: ${__supported}${__unsupported}")
```
tribal-tec: I remember removing this because of spam. It would be nice if there would be message(VERBOSE) which you can enable if you want it. You really think it is useful/needed?

rdumusc: I hadn't realised I was reintroducing the same thing you had removed... It prints only one line at the top of the cmake run, but if you don't like it we can think of something else to warn users if this executes multiple times. A message only if CMAKE_SOURCE_DIR != PROJECT_SOURCE_DIR for instance?

tribal-tec: Probably that's something more in line what you're looking for.

In general, the usefulness of the TestCPP11 is quite limited nowadays. All the features we check are present in all the compilers we use, and even if a certain feature is not present, I don't remember seeing any handling in that case.

CMake 3 solved that problem (finally), by specifying the required features on the target: http://www.cmake.org/cmake/help/v3.2/prop_gbl/CMAKE_CXX_KNOWN_FEATURES.html#prop_gbl:CMAKE_CXX_KNOWN_FEATURES

If we dare, we retire this TestCPP11 script, use the cmake feature for CMake 3 and CMake 2 does not perform any check.

rdumusc:
Updated

tribal-tec:
+1

PR #395, opened 2015-08-06
Exlude qrc files from check targets, unify exlude pattern handling

tribal-tec:


tribal-tec:
I needed this for DisplayCluster to pass the cppcheck run in finite time. It was stuck on one of the qrc files.

rdumusc:
+1

latk:
Errm, this change breaks my build. You have included this code:

```
if("${_clangcheck_lang}" MATCHES "CXX" AND
       NOT ${_clangcheck_loc} MATCHES ${add_clangcheck_EXCLUDE_PATTERN} AND
       ${_clangcheck_loc} MATCHES "\\.(cpp|cxx)$")
```

The problem is that `add_clangcheck_EXCLUDE_PATTERN` might not be defined, as is the case when using the function like `add_clangcheck(mytarget)`. In that case, the `if` construct sees the following arguments, which is invalid syntax:

```
if("CXX" MATCHES "CXX" AND
       NOT "/home/me/build/File.cpp" MATCHES AND
       "/home/me/build/File.cpp" MATCHES "\\.(cpp|cxx)$")
```

Remember that empty variables (i.e. variables that contain an empty string) do not survive interpolation – instead they are removed from the argument list, leading to this syntax. I know too little CMake to fix this myself, but perhaps splitting that rather lengthy conditional into multiple conditions and first testing whether the exclude pattern was set might be the start of a solution.

rdumusc:
@latk thanks for reporting this. Should be fixed with #398

PR #397, opened 2015-08-06
Use SUBPROJECT_TAG also for .gitexternals

eile:
(What Eyescale/config#48 really wanted to do)


GitExternal.cmake
```diff
@@ -40,6 +40,10 @@
 #    If set, a remote called 'user' is set up for github repositories, pointing
 #    to git@github.com:<user>/<project>. Also, this remote is used by default
 #    for 'git push'.
+#  GIT_EXTERNAL_TAG
```
rdumusc: Not an env variable. Move under GIT_EXTERNAL_VERBOSE.

rdumusc:
+1 otherwise

PR #401, opened 2015-08-11
More speedup (~30%)

tribal-tec:
- remove target hooks (slow cmake_parse_arguments), functionality
  provided by common_application() and common_library()
- remove *package* target (long and slow target list); futhermore,
  CPack is not subproject aware anyways


tribal-tec:
FYI @rdumusc @eile

rdumusc:
+1, nice cleanup.
Note that you did not remove the package target for subprojects as this one comes from CPack, included by CommonCPack (and CPackConfig), but the dput* targets.

CommonCPack.cmake
```diff
@@ -2,6 +2,12 @@
 # Copyright (c) 2012-2013 Stefan Eilemann <eile@eyescale.ch>
 # Info: http://www.itk.org/Wiki/CMake:Component_Install_With_CPack
 
+# No support for superproject packaging
```
rdumusc: subproject packaging. With this, the super-project can still be packaged.

eile:
+1

CommonCPack.cmake
tribal-tec: True. If the superproject includes CommonCPack. Not the case for config.bbp.

PR #403, opened 2015-08-14
Show output of ctest runs

eile:
Needed to see what's going on with https://bbpcode.epfl.ch/ci/job/viz.ISC.gerrit/342/build_type=Debug,platform=Ubuntu-14.04/console

Might revert later if it's too annoying.

### DIFF
--- a/CommonCPPCTest.cmake
+++ b/CommonCPPCTest.cmake
@@ -121,9 +121,14 @@ foreach(FILE ${TEST_FILES})
   endif()
 endforeach()
 
+set(__CONSOLE)
+if(CMAKE_VERSION VERSION_GREATER 3.1.99)
+  set(__CONSOLE USES_TERMINAL)
+endif()
+
 if(NOT TARGET ${PROJECT_NAME}-cpptests)
-  add_custom_target(${PROJECT_NAME}-cpptests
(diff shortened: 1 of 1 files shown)


CommonCPPCTest.cmake
```diff
@@ -121,9 +121,14 @@ foreach(FILE ${TEST_FILES})
   endif()
 endforeach()
 
+set(__CONSOLE)
+if(CMAKE_VERSION VERSION_GREATER 3.2.2)
```
tribal-tec: imo 3.2 is enough

tribal-tec:
+1

CommonCPPCTest.cmake
eile: Sure? I was it as new feature in 3.2.3, but maybe they have the same relnotes for all 3.2.x

rdumusc: Seems like 3.2 is enough: http://www.cmake.org/pipermail/cmake/2015-March/060082.html

rdumusc:
+1

CommonCPPCTest.cmake
eile: Ok, will change and then merge. Thanks.

PR #404, opened 2015-08-14
New CommonPackageConfig.cmake replaces PackageConfig.cmake

tribal-tec:
- targets are properly exported with CMake's export(EXPORT) feature
- CommonLibrary.cmake adds targets to export set
- generated ${PROJECT_NAME}Config.cmake handles build- and install-tree
  inclusion properly; only find dependencies if the targets are not present,
  which improves CMake run performance

Note that after the merge of this, all projects need to be updated accordingly.
I have all of them prepared, but this introduces another round of update- and
build ping-pong.


tribal-tec:
@rdumusc FYI

eile:
+1

rdumusc:
+1

PR #411, opened 2015-08-24
Fix for inclusion of CommonCPack without having used common_library.

hernando:
This is a legit use case, for example when only building a final
application.

### DIFF
--- a/CommonPackageConfig.cmake
+++ b/CommonPackageConfig.cmake
@@ -58,8 +58,8 @@ install(FILES "${PROJECT_BINARY_DIR}/${PROJECT_NAME}Config.cmake"
         DESTINATION ${CMAKE_MODULE_INSTALL_PATH}
 )
 
-# No export(EXPORT) in CMake 2
-if(CMAKE_MAJOR_VERSION LESS 3 OR NOT TARGET ${PROJECT_NAME})
+# No export(EXPORT) in CMake 2 or if common_library has not been used.
+if(CMAKE_MAJOR_VERSION LESS 3 OR NOT TARGET ${PROJECT_NAME}_ALIAS)
   return()
 endif()
 


CommonPackageConfig.cmake
```diff
@@ -63,6 +63,13 @@ if(CMAKE_MAJOR_VERSION LESS 3 OR NOT TARGET ${PROJECT_NAME})
   return()
 endif()
 
+# A project may want to include CommonCPack without having used common_library.
+# Returning in that case (Probably we don't event want to generate other
+# Config files since there's no library).
+if(NOT ${PROJECT_NAME}Targets)
```
tribal-tec: This does not work for me. Even if a project adds a target to the export set aka ${PROJECT_NAME}Targets, it early outs here. To my understanding, ${PROJECT_NAME}Targets is not a variable you can check like this.

hernando: OK, I'll find another way, but this use case needs to be sorted out. Otherwise CommonCPack.cmake is unusable for packages that only build applications.

tribal-tec: I just realized that in line 62 the check should be NOT TARGET ${PROJECT_NAME}_ALIAS. Then it would do what you want to do and it does what it was intended to do :)

hernando:
I did the change you commented.

hernando:
There's still a related subtle problem that causes spurious errors, for example in vmmlib. vmmlib is a header only library, so it doesn't set any ${PROJECT_NAME}_ALL_DEP_TARGETS, but on the other hand it sets ${PROJECT_NAME}_ALIAS.
If you try to configure vmmlib without Boost, vmmlib_ALL_DEP_TARGETS won't take any value because it was only set in unit test configuration, and without boost that's skipped.
This error has shown up while trying to configure Brion with a misconfiguration of Boost. Instead of complaining about Boost, CMake was complaining about vmmlib_ALL_DEP_TARGETS inside DoxygenRule.cmake, making the detection of the real problem almost impossible for a casual user. I'll submit a fix for this now.

PR #413, opened 2015-08-25
Fix #410: Restore API of check targets

### LINKED ISSUE #410: d33b31f breaks add_clangcheck() interface
Commit d33b31f7cb198f4dc87126060016e2899fc27daa changes the interface of `add_clangcheck()` – again. Previously, the syntax seems to have been

``` cmake
add_clangckeck(<target-name>
    [UNUSED_FUNCTIONS]
    [STYLE]
    [POSSIBLE_ERROR]
    [FAIL_ON_WARNINGS]
    [EXCLUDE_PATTERN <pattern>])
```

, but now it is

``` cmake
add_clangcheck(<target-name> <excluded-files-array>)
```

and documented as

``` cmake
add_clangcheck(<target-name>
    [UNUSED_FUNCTIONS]
    [STYLE]
    [POSSIBLE_ERROR]
    [FAIL_ON_WARNINGS])
```

This means that an invocation such as `add_clangeck(mytarget)` is no longer valid, and fails with the rather unhelpful error

``` none
CMake Error at <redacted>/CMakeLists.txt:24 (add_clangcheck):
  add_clangcheck Function invoked with incorrect arguments for function
  named: add_clangcheck
```

. Please either drop `clangcheckTargets` from the documented public interface in the README of this repository, or commit to a stable-ish, correctly documented, and backwards-compatible interface.

If backwards compatibility is not your thing, it might be worth marking known-good states as a release and encouraging users to clone that, rather than tracking `master` as suggested in the README:

> \[...]To use it, create a .gitexternals file in your project:
> 
> ``` none
> # -*- mode: cmake -*-
> # CMake/common https://github.com/Eyescale/CMake.git master
> ```
> Copy GitExternal from this repository to CMake/, and use it in your top-level CMakeLists.txt as follows:
> 
> ``` cmake
> list(APPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/CMake
>                               ${CMAKE_SOURCE_DIR}/CMake/common)
> include(GitExternal)
> include(Common)
> ```
>
> This will clone the latest version of this repository into your project at the beginning of the CMake run, and make all its features available. Run the 'update' target after the first successful cmake run to pin the git external to a SHA1 tag instead of a moving master tag.

(Which `update` target? Maybe my *real* problem is that I can't figure out how to do that…)

tribal-tec:


rdumusc:
now the code to get a list of files from a target is duplicated 4 time... make a reusable function?

tribal-tec:
Good point! Updated.

GetSourceFilesFromTarget.cmake
```diff
@@ -0,0 +1,18 @@
+# Copyright (c) 2015 Daniel.Nachbaur@epfl.ch
+#
+# Provide the macro get_source_files() to get all (C++) source files from the
```
rdumusc: function

CppcheckTargets.cmake
```diff
@@ -60,27 +62,44 @@ function(add_cppcheck _name _files)
...
-  list(FIND _input FAIL_ON_WARNINGS _fail_on_warn)
-  if("${_fail_on_warn}" GREATER "-1")
+  if(add_cppcheck_FAIL_ON_WARNINGS)
     list(APPEND CPPCHECK_FAIL_REGULAR_EXPRESSION
       ${CPPCHECK_WARN_REGULAR_EXPRESSION})
   endif()
 
+  set(_files ${add_cppcheck_FILES})
+  if(NOT _files)
+    get_target_property(_imported_target "${_name}" IMPORTED)
```
rdumusc: this check is done 3 times. Could it be moved inside get_source_files()?

tribal-tec: yes

rdumusc:
+1, looks very good now

tribal-tec:
Thx for the good review!

PR #416, opened 2015-08-28
Fix INTERFACE library for CMake 2, use proper replacement for CMake 3

tribal-tec:
FYI @ptoharia


rdumusc:
+1

ptoharia:
I know its already merged, I tested and works. Thanks! But I was using CMake 3.2.2

PR #418, opened 2015-09-03
Partially revert 3d951fd to fix OSX bundle generation

tribal-tec:


rdumusc:
+1. I'm not sure if this was removed for a particular reason, but if you have no problems now it's fine.

PR #421, opened 2015-10-15
Improvements to project directories

eile:


CommonDate.cmake
```diff
@@ -0,0 +1,12 @@
+# sets COMMON_DATE to today
+if(NOT COMMON_DATE)
+  if(WIN32)
+    execute_process(COMMAND "cmd" " /C date /T" OUTPUT_VARIABLE COMMON_DATE)
+    string(REGEX REPLACE "(..)/(..)/..(..).*" "\\1/\\2/\\3"
```
tribal-tec: move this outside the if(WIN32). It's the same for both branches

tribal-tec:
+1 otherwise

PR #422, opened 2015-10-20
Fix coverage report for multi library projects

tribal-tec:


rdumusc:
Hmm, what happens if ${PROJECT_SOURCE_DIR} contains subproject directories? Does it just gather more data, which is eventually filtered out at the next stage, or does this mix in the report?

tribal-tec:
Imo the gather step correlates the gdca files (the measured coverage) which are produced by unit tests against the source files. So if one runs tests from subdirectories, then yes, they appear in the same report. That's my understanding how this works.

rdumusc:
Yes, I wasn't sure any more, but that's why we had added this "filtering", otherwise you can't get clean per-project coverage reports for projects built "out-of-the-box". We need a different solution...

tribal-tec:
One solution could be that COMMON_SOURCE_DIR does not point by default to CMAKE_SOURCE_DIR so that subprojects are 'polluting' the source folder. Naturally, I would pick the parent folder of CMAKE_SOURCE_DIR, but that might not work with permissions?

rdumusc:
Not a bad idea, but we don't want to pollute the parent folder either (would be horrible to cleanup, might conflict with other local copies of the dependencies, etc...), so that would mean regrouping dependencies inside a subfolder of a parent folder (like ${ProjectName}-dependencies)

tribal-tec:
Solved by setting up multi library project properly (Servus with Qt lib).

PR #426, opened 2015-11-03
Fix git external update in super projects

eile:


tribal-tec:
Worked for me w/o this fix. How?

tribal-tec:
Should be COMMON_SOURCE_DIR imo, not CMAKE_SOURCE_DIR

eile:
> Worked for me w/o this fix. How?

You still had a CMake/common in the subproject.

eile:
> Should be COMMON_SOURCE_DIR imo, not CMAKE_SOURCE_DIR

Did you have to notice!? This is broken in so many places in this file.... fixing :@

eile:
Fixed.

hernando:
The commit is very broken in my opinion.
If I set COMMON_SOURCE_DIR now CMake/common goes into that path, which is fine, but there are two problems:
1. The biggest one. All projects are using ${CMAKE_SOURCE_DIR}/CMake/common and I don't see that you're updating that path.
2. The git clone command for CMake/common is failing badly. The code tries to set ${COMMON_SOURCE_DIR}/CMake as the working directory, but it doesn't exist.

hernando:
Another caveat to consider as a developer working on multiple projects and using a single source tree is that if you use separate build trees (not the usual case, but perfectly valid), every time you reconfigure a project it will checkout it's SHA1 in the common CMake/common. This is a potential pitfall for non expert users.

hernando:
There are more problems, see #428.

To me, this means that all projects need to be reviewed, because any project that is using this commit of CMake/common is not going to build out of the box.

eile:
> every time you reconfigure a project it will checkout it's SHA1 in the common CMake/common.

That's still the case? IIRC the SHA is only used for the initial clone, afterwards use the rebase target.

PR #428, opened 2015-11-04
Fix in git clone for CMake/common

hernando:
GitExternal is using a potentially non-existing directory to clone it
when COMMON_SOURCE_DIR is set, however the git command doesn't need
any specific working directory to do the job.

### DIFF
--- a/AppleCheckOpenGL.cmake
+++ b/AppleCheckOpenGL.cmake
@@ -11,10 +11,10 @@ function(apple_check_opengl Target)
 
   add_test(NAME ${Target}-AppleCheckOpenGL
     COMMAND ${CMAKE_COMMAND} -DAPPLE_CHECK_OPENGL_FILE="$<TARGET_FILE:${Target}>"
-            -P ${CMAKE_SOURCE_DIR}/CMake/common/AppleCheckOpenGL.cmake)
+            -P ${COMMON_SOURCE_DIR}/CMake/common/AppleCheckOpenGL.cmake)
   add_custom_target(${Target}-AppleCheckOpenGL
     COMMAND ${CMAKE_COMMAND} -DAPPLE_CHECK_OPENGL_FILE="$<TARGET_FILE:${Target}>"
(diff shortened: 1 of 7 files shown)


eile:
Oh dear, what a can of worms I found. I think it's better if we leave .gitexternals in CMAKE_SOURCE_DIR?

hernando:
It's going to be simpler.

What was the reason for the change you made?

eile:
We half-moved (at least per doc) it to COMMON_SOURCE_DIR, and I was just trying to fix/finish this.

hernando:
per doc? I'm not following.

In this PR I've searched for all the appearances of CMake/common in the scripts and change them. This works, but it requires to adapt projects to use COMMON_SOURCE_DIR/CMake/common as the CMake module directory.

eile:
https://github.com/Eyescale/CMake/pull/426/files#diff-342bd6e1b740e0a1d015c07a635ea2afL154 seemed to indicate that git externals already should go to CMAKE_SOURCE_DIR. Maybe I misread.

hernando:
OK, you mixed .gitexternals and git externals in your comments and that confused me.
I'm not in favor of moving git externals out of COMMON_SOURCE_DIR. The current code was considering that if you use COMMON_SOURCE_DIR it's because you know what you're doing. Under this assumption, there's no need to use it as the default path for clone operations. With #426  you didn't provide any benefit for regular users, and broke the build for power users because CMake/common has been shifted to COMMON_SOURCE_DIR. So the solutions are:

- Leave it as it was and don't use COMMON_SOURCE_DIR if you're not willing to clone the projects manually.
- Finish the change and adapt all projects to the CMake/common shift.
- Handle CMake/common differently in GitExternal and always clone it under CMAKE_SOURCE_DIR.

I can't think of any other alternatives.

eile:
git subprojects should be cloned under COMMON_SOURCE_DIR
.gitexternals should be cloned under CMAKE_SOURCE_DIR (for now to not change all projects)

hernando:
Agreed. I was confusing git externals with git subprojects.

PR #430, opened 2015-11-04
Make empty -doxycopy target warning visible when run.

hernando:


bilgili:
+1

DoxygenRule.cmake
```diff
@@ -186,11 +186,17 @@ if(GIT_DOCUMENTATION_REPO)
       COMMENT "Copying ${PROJECT_NAME} API documentation to ${GIT_DOCUMENTATION_DIR}"
       DEPENDS ${PROJECT_NAME}-doxygen VERBATIM)
   else()
+    # Having a command while force this target to run always when invoked. That
```
rdumusc: while -> will

DoxygenRule.cmake
```diff
@@ -186,11 +186,17 @@ if(GIT_DOCUMENTATION_REPO)
...
+    # makes the Makefile to print the comment. The text will be also visible
+    # in ninja.
     add_custom_target(${PROJECT_NAME}-doxycopy
+      COMMAND ${CMAKE_COMMAND} -E echo ""
       COMMENT "doxycopy target not available, missing ${_GIT_DOC_SRC_DIR}")
```
rdumusc: What about merging COMMAND -E echo and COMMENT? Would it do the same..?

DoxygenRule.cmake
hernando: I'm becoming dyslexic.. the more I write in English the more I made this word substitution mistakes.

rdumusc:
I'm not against this change, but I still wonder under which conditions this message is not printed. As far as I can remember this has always worked for me:
make DisplayCluster-doxycopy
Scanning dependencies of target DisplayCluster-doxycopy
doxycopy target not available, missing /home/dumusc/src/stable/DisplayCluster/../BlueBrain
Built target DisplayCluster-doxycopy

DoxygenRule.cmake
hernando: Not sure. It looks nicer like this with the Makefile generator because it stands out with blue ink.

hernando:
It's a CMake 2 vs 3 version. With CMake 2 it gets printed, not with 3. Probably CMake 3 detects the empty target and generates less code.

rdumusc:
hah! OK, +1 for me

PR #431, opened 2015-11-06
Updated Coverage.cmake to include full source tree

mgeplf:
* before, ${PROJECT_SOURCE_DIR}/${PROJECT_INCLUDE_NAME} was being
  used as one of the 'gathering' directories, made this more general,
  so source trees with different layouts will work
* also fixed a typo

### DIFF
--- a/Coverage.cmake
+++ b/Coverage.cmake
@@ -8,7 +8,7 @@
 #
 # Input variables:
 # * LCOV_EXCLUDE Extra files to exclude from the coverage report
-# * COVERAGE_LIMITS Optional genhml flags to tweak the color codes of the report
+# * COVERAGE_LIMITS Optional genhtml flags to tweak the color codes of the report
 #
 # Targets generated:
 # * lcov-clean_${PROJECT_NAME} for internal use - clean before running cpptests
@@ -74,8 +74,7 @@ function(add_coverage_targets TEST_TARGET)
 
   if(NOT TARGET lcov-gather_${PROJECT_NAME})
     add_custom_target(lcov-gather_${PROJECT_NAME}
-      COMMAND ${LCOV} -q --capture --directory . --no-external
-        --directory ${PROJECT_SOURCE_DIR}/${PROJECT_INCLUDE_NAME}
+       COMMAND ${LCOV} --capture --no-external --directory . --directory ${CMAKE_SOURCE_DIR}
         --output-file lcov.info
       COMMENT "Capturing code coverage counters"
       WORKING_DIRECTORY ${PROJECT_BINARY_DIR})


rdumusc:
I understand your motivation, because the current situation is a bit too restrictive,
but I'm afraid we can't remove this filtering so easily. Without it, a typical project built out-of-the-box will include the coverage report for all its .gitsubprojects (which are cloned in PROJECT_SOURCE_DIR), and this is not what we want.
Maybe we could let users provide a list of sub-directories to be included in the report? A similiar option existed in the past but was removed here, maybe this can be somehow reused: https://github.com/Eyescale/CMake/commit/b9fbac3ab9b3ac40bac8bf08b401ce49f68457a6

mgeplf:
Would it be possible to have 3rd party libs cloned to a different directory, instead the same one as the PROJECT_SOURCE? Then that directory can be explicitly filtered out, b/c we know it exists.  PROJECT_INCLUDE_NAME isn't a standard cmake variable, which means that this module could be more re-useable if there wasn't this dependency.

I'm still learning this build system, so please forgive me for being naive.  I suspect that such a change would be difficult.  Can you point me to a repo that relies on the .gitsubprojects and coverage so I can play with it?

rdumusc:
There is exactly such a feature which can be enabled by setting COMMON_SOURCE_DIR (look for its usage). It is not fully production-ready however, and unlikely to become the default any time soon.

Sure, thanks for trying to help making this code more reusable :-) Pretty much any of our projects uses the .gitexternals feature. Take a look at https://github.com/BlueBrain/DisplayCluster and https://github.com/BlueBrain/Deflect for instance, with the existing coverage reports here:
http://bluebrain.github.io/

mgeplf:
I did a build of viz-stable w/ -DCOMMON_SOURCE_DIR=`pwd`/../3rdParty, and it seems like it worked properly.  What other testing could we do to make this production ready?  I'm personally a fan of not adding too much to the cmake project directory, so I would be happier if all the dependencies were under one directory versus them being interspersed with the rest of the project directory.

rdumusc:
Yes it works for most use cases, and you're welcome to use it if you like it better!
I am not sure what problems still remain, but even if it works perfectly, and even if it can arguably be considered better practice to not alter the source folder, the reality is that we are not going to enforce placing subproject sources outside the source folder overnight - if we do it at all, because most people are happy with the current situation and will see no reason for yet another change of practice - so I don't think we can rely on that to solve the problem of the coverage report

eile:
Plus in some cases (Jenkins) you might not be able to place them outside.

mgeplf:
Sorry, I wasn't being clear, it's still in the root of the project, just that the subprojects are checked out into one place:
```
|-- 3rdParty
|   |-- dep0
|   `-- dep1
|-- awesome_project
`-- CMakeLists.txt
```
vs:
```
|-- awesome_project
|-- CMakeLists.txt
|-- dep0
`-- dep1
```

Anyways, not important.  I'll close this up, and add what I need to the SubProject part, like we discussed, when I have some time.

PR #433, opened 2015-11-10
Let users scroll the documentation TOC when it doesn't fit on the screen

rdumusc:
@hernando FYI


rdumusc:
Sorry for the multiple iterations, but overflow-y: auto; is the correct property

PR #434, opened 2015-11-13
Remove -DBOOST_TEST_DYN_LINK from all translation units

mgeplf:
* Above define was being set on all translation units,
  this is only necessary for boost tests, and should be
  defined only when required

### DIFF
--- a/CommonCPPCTest.cmake
+++ b/CommonCPPCTest.cmake
@@ -48,9 +48,11 @@ list(SORT TEST_FILES)
 set(ALL_CPP_TESTS)
 set(ALL_CPP_PERF_TESTS)
 
-set(TEST_ENABLE_BOOST_HEADER "Add -DBOOST_TEST_DYN_LINK to tests" CACHE BOOL ON)
 # backwards compat: generate main() for unit tests
 #  should really #define BOOST_TEST_DYN_LINK if using boost
+if(NOT DEFINED TEST_ENABLE_BOOST_HEADER)
+  set(TEST_ENABLE_BOOST_HEADER ON)
+endif()
 if(NOT Boost_USE_STATIC_LIBS AND TEST_ENABLE_BOOST_HEADER)
   add_definitions(-DBOOST_TEST_DYN_LINK)
 endif()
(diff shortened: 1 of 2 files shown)


eile:
Can you add in in CommonCPPCTest then?

juanchopanza:
I would prefer to add this to my tests explicitly. A) I want the code for my tests to work independent of this cmake config and B) I may use a different testing framework which doesn't need this define. Defining the macro hurts in case A)

    error: "BOOST_TEST_DYN_LINK" redefined [-Werror]

and probably not in case B)

mgeplf:
I'll add it in CommonCPPCTest with an option to turn it off.  Can I have a warning saying 'You should add this to your source file'?

CommonCPPCTest.cmake
```diff
@@ -47,6 +48,13 @@ list(SORT TEST_FILES)
 set(ALL_CPP_TESTS)
 set(ALL_CPP_PERF_TESTS)
 
+set(TEST_ENABLE_BOOST_HEADER "Add -DBOOST_TEST_DYN_LINK to tests" CACHE BOOL ON)
```
mgeplf: Do you want this to be an advanced option?

eile: Making it an option is a bit pointless, since in superproject builds each project needs a different setting. Why not have it a simple, non-cache variable projects set in their tests/CMakeLists.txt?

eile:
> warning saying 'You should add this to your source file'

I feel no, since we're working hard on not being noisy.

@rdumusc, @tribal-tec ?

CommonCPPCTest.cmake
mgeplf: Fair enough.  So the idea would be to:

#keep legacy behavior
if(NOT DEFINED TEST_ENABLE_BOOST_HEADER)
  set(TEST_ENABLE_BOOST_HEADER ON)
endif()
instead?

mgeplf:
I prefer silence.  It's a simple grep to find which files are missing this.

tribal-tec:
CommonCPPTest is fine. In the end, to my knowledge it's only needed on Windows to fix a linker error wrt missing main for each boost unit test. It could be automatically added if boost unit test library is found or linked to the executable? Then the user-option would not be needed as a user probably doesn't know or care about this define.

PR #437, opened 2015-11-13
Have SubProject return a list of directories

mgeplf:
* SUBPROJECT_PATHS is a list of directories of the subprojects,
  it is percolated up such that it can be used for exclusion in
  coverage, for example

### DIFF
--- a/SubProject.cmake
+++ b/SubProject.cmake
@@ -229,10 +229,9 @@ if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/.gitsubprojects")
         "else()\n"
         "  file(APPEND .gitsubprojects \"git_subproject(${__subproject})\n\")\n"
         "endif()\n")
-      set(__subproject_paths "${__subproject_paths} ${__subproject_dir}")
+        list(APPEND __subproject_paths ${__subproject_dir})
     endforeach()
 
-    separate_arguments(__subproject_paths)
     list(REMOVE_DUPLICATES __subproject_paths)
     set_property(GLOBAL PROPERTY SUBPROJECT_PATHS ${__subproject_paths})
 
--- a/SubProject.cmake
+++ b/SubProject.cmake
@@ -205,8 +205,8 @@ if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/.gitsubprojects")
 
   set(__subprojects) # appended on each git_subproject invocation
   include(.gitsubprojects)
-
   if(__subprojects)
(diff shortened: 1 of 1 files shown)


SubProject.cmake
```diff
@@ -225,8 +229,13 @@ if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/.gitsubprojects")
         "else()\n"
         "  file(APPEND .gitsubprojects \"git_subproject(${__subproject})\n\")\n"
         "endif()\n")
+     set(SUBPROJECT_PATHS "${SUBPROJECT_PATHS} ${__subproject_dir}" 
+       CACHE INTERNAL SUBPROJECT_PATHS)
```
mgeplf: Not a huge fan of namespace pollution: if there's a more idiomatic way of doing this, let me know, and I'll change it

eile: A slight better way is to set a global property, at least it's not kept between cmake runs:
```
    set_property(GLOBAL APPEND PROPERTY
                 ${PROJECT_NAME}_ALL_DEP_TARGETS ${LibName})
```

mgeplf: Ahh, interesting, I will fix it Monday.  Thanks for the tip, and the patience.

SubProject.cmake
```diff
@@ -225,8 +229,13 @@ if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/.gitsubprojects")
         "else()\n"
         "  file(APPEND .gitsubprojects \"git_subproject(${__subproject})\n\")\n"
         "endif()\n")
+      set(__subproject_paths "${__subproject_paths} ${__subproject_dir}")
```
eile: ```set_property(GLOBAL APPEND ...)``` instead of the two added lines above?

mgeplf: If I remember properly, I think I tried that, but I wanted to de-duplicate the list, so I have to change it into a variable at some point and thus it ended up being the same, just slightly re-arranged.

eile: Ah :)

Ok, +1 from me then.

rdumusc: replace
set(__subproject_paths "${__subproject_paths} ${__subproject_dir}")
with
list(APPEND __subproject_paths ${__subproject_dir})
and remove separate_arguments(__subproject_paths)?

rdumusc:
ping

mgeplf:
Sorry, missed that.   I'll try and find time today to try that and test it.  Seems reasonable.

mgeplf:
@rdumusc 

Made the changes, thanks for the tip.

PR #438, opened 2015-11-16
default OSX build is only x86_64

mgeplf:


tribal-tec:
+1. @eile might not like it :)

eile:
We discussed this offline, I can add the correct OS X archs manually.

PR #440, opened 2015-11-20
Complement Eyescale/Lunchbox#241

eile:


cpp/test.h
```diff
@@ -89,15 +89,15 @@ class Watchdog : public lunchbox::Thread
             lunchbox::Thread::setName( "Watchdog" );
 #ifdef TEST_RUNTIME
             lunchbox::sleep( TEST_RUNTIME * 1000 );
-            TESTINFO( false,
-                      "Watchdog triggered - " << _name <<
-                      " did not terminate within " << TEST_RUNTIME << "s" );
+            std::cerr << "Watchdog triggered - " << _name
```
tribal-tec: why changed?

tribal-tec:
+1 otherwise

cpp/test.h
eile: Because it was just a hack to get the abort for free, which is now called explicitly.

PR #443, opened 2015-12-07
Allow for more fine-grained flag setting

mgeplf:
* Break out compiler identification into its own file
* allow for disabling changes to CMAKE_CXX_FLAGS w/ COMMON_DISABLE

### DIFF
--- a/Compiler.cmake
+++ b/Compiler.cmake
@@ -1,7 +1,7 @@
 # Copyright (c) 2012-2014 Fabien Delalondre <fabien.delalondre@epfl.ch>
 #                         Stefan.Eilemann@epfl.ch
 #
-# Sets compiler optimization, definition and warnings according to
+# Sets compiler optimization, and warnings according to
 # chosen compiler. Supported compilers are XL, Intel, Clang, gcc (4.4
 # or later) and Visual Studio (2008 or later).
 #
@@ -11,6 +11,7 @@
 # Input Variables
 # * COMMON_MINIMUM_GCC_VERSION check for a minimum gcc version, default 4.4
 # * COMMON_USE_CXX03 When set, do not enable C++11 language features
+# * COMMON_DISABLE When set, do not set any flags
 #
 # Output Variables
 # * GCC_COMPILER_VERSION The compiler version if gcc is used
@@ -22,10 +23,6 @@
 # * COMMON_CXX_FLAGS_DEBUG The common flags for C++ compiler in Debug build
 # * COMMON_CXX_FLAGS_RELWITHDEBINFO The common flags for C++ compiler in RelWithDebInfo build
 # * COMMON_CXX_FLAGS_RELEASE The common flags for C++ compiler in Release build
-# * C_DIALECT_OPT_C89    Compiler flag to select C89 C dialect
-# * C_DIALECT_OPT_C89EXT Compiler flag to select C89 C dialect with extensions
-# * C_DIALECT_OPT_C99    Compiler flag to select C99 C dialect
-# * C_DIALECT_OPT_C99EXT Compiler flag to select C99 C dialect with extensions
 
 # OPT: necessary only once, included by Common.cmake
 if(COMPILER_DONE)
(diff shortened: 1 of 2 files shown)


mgeplf:
This is an intermediate commit to allow us to do different things with our flags more easily

Compiler.cmake
```diff
@@ -33,34 +30,19 @@ if(COMPILER_DONE)
...
 set(COMPILER_DONE ON)
 
-include(TestBigEndian)
-test_big_endian(BIGENDIAN)
-
-if(BIGENDIAN)
-  add_definitions(-DCOMMON_BIGENDIAN)
-else()
-  add_definitions(-DCOMMON_LITTLEENDIAN)
+if(COMMON_DISABLE)
```
mgeplf: so we can disable flags being set

Compiler.cmake
```diff
@@ -33,34 +30,19 @@ if(COMPILER_DONE)
...
+include(CompilerIdentification)
 
 if(NOT COMMON_MINIMUM_GCC_VERSION)
   set(COMMON_MINIMUM_GCC_VERSION 4.4)
 endif()
 
-option(ENABLE_WARN_DEPRECATED "Enable deprecation warnings" ON)
 option(ENABLE_CXX11_STDLIB "Enable C++11 stdlib" OFF)
 
+option(ENABLE_WARN_DEPRECATED "Enable deprecation warnings" ON)
```
mgeplf: just moved this closer to where it's used

Compiler.cmake
```diff
@@ -157,14 +131,8 @@ elseif(CMAKE_COMPILER_IS_XLCXX)
   set(CMAKE_C_FLAGS_RELEASE ${COMMON_C_FLAGS_RELEASE})
```
mgeplf: Can this be removed, and set, so CMAKE_C(XX)_FLAGS(_RELEASE|...) are set all in the same place?

Compiler.cmake
```diff
@@ -1,7 +1,7 @@
 # Copyright (c) 2012-2014 Fabien Delalondre <fabien.delalondre@epfl.ch>
 #                         Stefan.Eilemann@epfl.ch
 #
-# Sets compiler optimization, definition and warnings according to
```
eile: should have been 'defines'. Can you re-add it?

Compiler.cmake
```diff
@@ -11,6 +11,7 @@
 # Input Variables
 # * COMMON_MINIMUM_GCC_VERSION check for a minimum gcc version, default 4.4
 # * COMMON_USE_CXX03 When set, do not enable C++11 language features
+# * COMMON_DISABLE When set, do not set any flags
```
eile: There is an underlying bigger problem: In dependent projects, some flags have to be the same. This is the reason that for example COMMON_USE_CXX03 is a global option.

I would rather find a way to optionally disable 'safe' flags such as some warnings on a per-project level.

Can you give a list of flags which are problematic for you?

Compiler.cmake
eile: Looks like a leftover to me. Check with HPC team?

Compiler.cmake
mgeplf: Yeah, agreed about the problem.  CMake 3 seems to have target_compile_features and target_compile_definitions which seems like the future.

I'm generally fine w/ your flags, but Juan was hoping for something different.  I'll see if I can get him to comment

Compiler.cmake
mgeplf: Who do you recommend I ask?

eile: git blame says @delalond

Compiler.cmake
juanchopanza: If this is to be used generally, then I think it is important that arbitrary flags can be set. At the top level. If we need to build a consistent set of libraries for BBP consumption, then we can agree to use compatible flags, and even have a bbp-official-cpp-flags cmake module. Regarding flag compatibility between dependencies, how do we check / ensure that in the current code WRT installed dependencies?

mgeplf: Flags used are:
-Wall -Wextra -Wuninitialized -Winit-self -Woverloaded-virtual -Werror=pedantic -pedantic -pedantic-errors -Wno-long-long -Wno-missing-field-initializers -std=c++98

eile: @juanchopanza: This file is the attempt for an 'bbp-official-cpp-flags cmake module'. So far the uptake has been medium, therefore I expect some haggling from you (and others).

We do not check for compatibility in flags yet. Not sure how this can be done.

eile: @mgeplf: Default is -std=c++11 for gcc 4.8 (which is the swag-ed compiler on Intel, if that still means anything)

mgeplf: @eile 
w/ cmake 3, there is https://cmake.org/cmake/help/v3.0/module/CheckCXXCompilerFlag.html for checking flaga compatibility.  I haven't looked if this exists in the cmake 2.x series.

I will probably come down sometime so we can discuss this face-to-face.  @juanchopanza let me know if you interested in being there.

juanchopanza: @mgeplf Yes, I would definitely be interested in being there.

eile: @mgeplf : This is one of the many new things in CMake 3. Unfortunately I'm not sure we can abandon CMake 2.8 yet.

F2F sounds good - Catch us between Lunch and the 2pm presentation? Or let's have lunch together>

mgeplf: Both Juan and I have to do all hands stuff today, so time is tight.  Maybe another day this week?

juanchopanza: Or after the all-hands meeting?

PR #446, opened 2015-12-14
Coverage.cmake fixes

tribal-tec:
* Fix missing coverage report for multiple project directories
* Rename targets to match PROJECT_NAME-target syntax
* Exclude only generated files from COMMON_GENERATED_FILES property,
  not everything from PROJECT_BINARY_DIR


tribal-tec:
FYI @rdumusc

PR #448, opened 2015-12-17
Replace + by P in package names for define.h generation

favreau:


eile:
Can you squash the three commits, cancelling each other out?

favreau:
Done :-)

PR #449, opened 2015-12-17
added Gcovr coverage tool support

mgeplf:


CoverageGcovr.cmake
```diff
@@ -0,0 +1,66 @@
...
+
+if(ENABLE_GCOVR)
+  if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_COMPILER_IS_CLANG)
+     find_program(GCOVR gcovr)
+     if(NOT GCOVR)
+       message(FATAL_ERROR "No code coverage report, ${COVERAGE_MISSING}")
+     endif()
+
+     set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} ${__profiling_flags}")
+     set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} ${__profiling_flags}")
```
eile: Only in debug builds?

juanchopanza:
Is there something else that ensures that coverage only happens in DEBUG build type? Otherwise shouldn't this modify more general variables (possibly `CMAKE_CXX_FLAGS` etc. although I am not sure how this gets handled here.)

eile:
Is there a reason to limit coverage to debug builds?

juanchopanza:
Probably for reliability with `gcov`, coverage stats should be obtained without optimization. See

https://gcc.gnu.org/onlinedocs/gcc/Gcov-and-Optimization.html

eile:
Ok, then looks good for me. @mgeplf ready to merge?

mgeplf:
Fine w/ me

PR #450, opened 2015-12-18
Replace + symbol by P in package names

favreau:


eile:
Fixed by #451 in a nicer way.

PR #453, opened 2015-12-23
split Compiler.cmake to separate compiler ident from setting options

mgeplf:
* new file: CompilerIdentification.cmake, takes care of setting
   - compiler names (CMAKE_COMPILER_IS_...), and the C and C++ dialect
     options: (C_DIALECT_OPT_..., CXX_DIALECT_...) which cleans up the
     logic in the Compiler.cmake to mainly deal with flags

Note: with g++, I used std=c++03 is an alias to c++98, in the
different versions of source I've looked at, hence the term
'CXX_DIALECT_PRE_11' is used to signify both.

Investigation:
```
GCC:
   #Default RH6.7
   $ g++ -dumpversion
   4.4.7
   $ echo | g++ -std=c++03 -x c++ -E -
   cc1plus: error: unrecognized command line option "-std=c++03"
   $ echo | g++ -std=c++98 -x c++ -E -

   #default Ubuntu LTS12.04
   $ g++ -dumpversion
   4.6
   $ echo | g++ -std=c++03 -x c++ -E -
   $ echo | g++ -std=c++98 -x c++ -E -
   # Note: I think this is -std=c++03 is an alias for -std=c++98
   #https://gcc.gnu.org/git/?p=gcc.git;a=blob;f=gcc/c-family/c.opt;h=4c4727f31c5719b6fbf2db0b1e9fdb497b86b142;hb=31839765ee7d53766a69f418e9d7a4ed750de4e9#l1143
   #Thanks to Juan for mentioning something like this may exist

   # from BBP/viz/env/2015.10.22
   $ g++ -dumpversion
   4.8.3
   $ echo | g++ -std=c++03 -x c++ -E -
   gevaert@bbpviz1 ~/private/src/CMake (master)$ echo | g++ -std=c++98 -x c++ -E -
   #alias for c++98:
   #https://gcc.gnu.org/git/?p=gcc.git;a=blob;f=gcc/c-family/c.opt;h=4da80b0edf558d2ab7b2ce740d0fca4fc4901929;hb=6bbf0dec66c0e719b06cd2fe67559fda6df09000#l1367

ICC:
   # from module load intel/icomposer-2015.0.090
   icpc -help
      snip>>
      -std=<std>
      c++11 enable C++11 experimental support for C++ programs
      c++0x same as c++11
      gnu++98 conforms to 1998 ISO C++ standard plus GNU extensions
      gnu++11 conforms to 2011 ISO C++ standard plus GNU extensions
      gnu++0x same as gnu++11
      snip>>
```

### DIFF
--- a/Compiler.cmake
+++ b/Compiler.cmake
@@ -56,7 +56,7 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_COMPILER_IS_CLANG)
 
   if(CMAKE_COMPILER_IS_CLANG)
     set(COMMON_C_FLAGS
-      "${COMMON_C_FLAGS} -Qunused-arguments -ferror-limit=6 -ftemplate-depth-1024 -Wheader-hygiene")
+      "${COMMON_C_FLAGS} -Qunused-arguments -ferror-limit=5 -ftemplate-depth-1024 -Wheader-hygiene")
     set(COMMON_CXX11_STDLIB "-stdlib=libc++")
   else()
     if(GCC_COMPILER_VERSION VERSION_LESS COMMON_MINIMUM_GCC_VERSION)
--- a/Compiler.cmake
+++ b/Compiler.cmake
@@ -22,11 +22,6 @@
 # * COMMON_CXX_FLAGS_DEBUG The common flags for C++ compiler in Debug build
 # * COMMON_CXX_FLAGS_RELWITHDEBINFO The common flags for C++ compiler in RelWithDebInfo build
 # * COMMON_CXX_FLAGS_RELEASE The common flags for C++ compiler in Release build
-# * C_DIALECT_OPT_C89    Compiler flag to select C89 C dialect
-# * C_DIALECT_OPT_C89EXT Compiler flag to select C89 C dialect with extensions
-# * C_DIALECT_OPT_C99    Compiler flag to select C99 C dialect
-# * C_DIALECT_OPT_C99EXT Compiler flag to select C99 C dialect with extensions
-# * COMMON_USE_CXX03 Set if the compiler does only support C++03
 
 # OPT: necessary only once, included by Common.cmake
 if(COMPILER_DONE)
@@ -34,30 +29,7 @@ if(COMPILER_DONE)
 endif()
 set(COMPILER_DONE ON)
 
-include(TestBigEndian)
-test_big_endian(BIGENDIAN)
-
-if(BIGENDIAN)
-  add_definitions(-DCOMMON_BIGENDIAN)
-else()
-  add_definitions(-DCOMMON_LITTLEENDIAN)
-endif()
-
-# Compiler name
-if(CMAKE_CXX_COMPILER_ID STREQUAL "XL")
-  set(CMAKE_COMPILER_IS_XLCXX ON)
-elseif(CMAKE_CXX_COMPILER_ID STREQUAL "Intel")
-  set(CMAKE_COMPILER_IS_INTEL ON)
-elseif(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
-  set(CMAKE_COMPILER_IS_CLANG ON)
-elseif(CMAKE_COMPILER_IS_GNUCXX)
-  set(CMAKE_COMPILER_IS_GNUCXX_PURE ON)
-endif()
-# use MSVC for Visual Studio
-
-if(NOT COMMON_MINIMUM_GCC_VERSION)
-  set(COMMON_MINIMUM_GCC_VERSION 4.4)
-endif()
+include(${CMAKE_CURRENT_LIST_DIR}/CompilerIdentification.cmake)
 
 option(ENABLE_WARN_DEPRECATED "Enable deprecation warnings" ON)
 option(ENABLE_CXX11_STDLIB "Enable C++11 stdlib" OFF)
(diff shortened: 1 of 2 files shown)


mgeplf:
Note, this needs to be tested on your guys end, in case any of the behavior has changed.  I've tried to tread lightly

Compiler.cmake
```diff
@@ -101,27 +74,13 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_COMPILER_IS_CLANG)
     # use C++03 std and stdlib, which is the default used by all
     # software, including all MacPorts.
   elseif(NOT COMMON_USE_CXX03)
-    if(CMAKE_COMPILER_IS_CLANG)
```
mgeplf: I think the benefit comes out here, as I think the new method more clearly shows the intent

Compiler.cmake
```diff
@@ -101,27 +74,13 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_COMPILER_IS_CLANG)
...
-      if(GCC_COMPILER_VERSION VERSION_LESS 4.7)
-        set(COMMON_CXXSTD_FLAGS "-std=c++0x")
-      else()
-        if(GCC_COMPILER_VERSION VERSION_LESS 4.8)
-          # http://stackoverflow.com/questions/4438084
-          add_definitions(-D_GLIBCXX_USE_NANOSLEEP)
-        endif()
-        set(COMMON_CXXSTD_FLAGS "-std=c++11")
-      endif()
+    set(COMMON_CXXSTD_FLAGS ${CXX_DIALECT_11})
```
juanchopanza: Note that this logic will have to change if we ever want to select C++14. Shouldn't we think of setting C++03 (or pre-C++11) flags if `COMMON_USE_CXX03` is set instead setting c++11 if it isn't set?

mgeplf: I agree, we should probably do something more explicit like that.  Having a COMMON_USE_CXX11 would probably be a good idea, as would removing the 'if(APPLE AND OSX_VERSION VERSION_LESS 10.9)' if possible.

Anyways, one step at a time

CompilerIdentification.cmake
```diff
@@ -0,0 +1,79 @@
...
+
+  set(CXX_DIALECT_PRE_11 "-std=c++98")
+
+  set(CXX_DIALECT_11 "-std=c++11")
+  if(GCC_COMPILER_VERSION VERSION_LESS 4.7)
+    set(CXX_DIALECT_11 "-std=c++0x")
+  endif()
+
+  if(CMAKE_COMPILER_IS_GNUCXX_PURE AND GCC_COMPILER_VERSION VERSION_LESS 4.5)
+    set(COMMON_USE_CXX03 ON)
```
eile: Add ```set(CXX_DIALECT_PRE_11 "")```?

CompilerIdentification.cmake
```diff
@@ -0,0 +1,79 @@
...
+  set(CXX_DIALECT_PRE_11 "-std=gnu++98")
+
+  set(C_DIALECT_OPT_C89 "-std=c89")
+  set(C_DIALECT_OPT_C99 "-std=c99")
+  set(C_DIALECT_OPT_C89EXT "-std=gnu89")
+  set(C_DIALECT_OPT_C99EXT "-std=gnu99")
+
+elseif(CMAKE_COMPILER_IS_XLCXX)
+  set(COMMON_USE_CXX03 ON)
+  set(CXX_DIALECT_PRE_11 "")
```
eile: Add ```set(CXX_DIALECT_11 "")```?

Compiler.cmake
```diff
@@ -66,22 +38,25 @@ if(ENABLE_WARN_DEPRECATED)
...
   endif()
+
   if(GCC_COMPILER_VERSION VERSION_GREATER 4.1)
     set(COMMON_C_FLAGS "${COMMON_C_FLAGS} -Wshadow")
   endif()
+
   if(CMAKE_COMPILER_IS_CLANG)
     set(COMMON_C_FLAGS
-      "${COMMON_C_FLAGS} -Qunused-arguments -ferror-limit=5 -ftemplate-depth-1024 -Wheader-hygiene")
+      "${COMMON_C_FLAGS} -Qunused-arguments -ferror-limit=6 -ftemplate-depth-1024 -Wheader-hygiene")
```
eile: ?

Compiler.cmake
eile: Agreed

eile:
+1 otherwise

Compiler.cmake
mgeplf: sad face

CompilerIdentification.cmake
mgeplf: I can do that.  

Is there value in a message being printed to the user that c++11 isn't supported?  The way it is now, if someone uses a compiler that doesn't have -std=c++11, the compiler will fail early saying it's not supported, which I think is valuable: ie, you tried to use something that won't work.

juanchopanza: I would say to not fail early would be a problem in this case. Failure with an easy to understand message would be the best option.

rdumusc: I agree in principle, but not all projects needs C++11 (or C++14 sooner or later) and there is currently no input variable to determine what is required.
So I can see two options to add this check :
- each project declared a minimum required C++ version in its top-level CMakeLists as: set(${PROJECT}_CXX_DIALECT 11) and include(Compiler) does the check if VERSION_LOWER.
- leave the responsibility to each individual project to check a CXX_DIALECT variable exported by this file

juanchopanza: Well, is you ask for C++11 and it isn't there, it should fail, period. You don't have to ask if you don't care.

rdumusc: yes

CompilerIdentification.cmake
rdumusc: +1 for consistency

rdumusc:
+1 good for me, this change makes things cleaner. Just address the last detail and we can merge.

CompilerIdentification.cmake
mgeplf: Is this the detail you want changed?  I think this should issue a warning saying that xlc doesn't have cxx11

rdumusc: yes for the detail. Well, I think that there is no reason to print a warning all the time; I expect most people using XLC are aware of its features and don't want the extra verbosity. As discussed above, what we could have instead is a fatal error when trying to configure a project that explicitly requires C++11 support, and this for any compiler, but that's a new feature.

eile: Plus, it will fail on the first C++11 code anyways.

rdumusc:
sorry, didn't see the update. all good now!

mgeplf:
No problem, thanks for the input.

PR #454, opened 2015-12-23
First attempt at adding run-time sanitizers

mgeplf:
* Add sanitizers for clang and gcc

### DIFF
--- a/Sanitizer.cmake
+++ b/Sanitizer.cmake
@@ -1,0 +1,48 @@
+# Used to turn on certain sanitizers, if available:
+# 
+# clang:
+#    -dumpcompiler: 4.2.1
+#       address
+#       thread
+#       undefined
+# gcc:
+#    -dumpcompiler: 4.8.3
+#       address
+#       thread
+#
+# Note: This modifies CMAKE_CXX_FLAGS and CMAKE_EXE_LINKER_FLAGS
+
+option(SANITIZER "Sanitizer: one of address/thread/undefined, depending on compiler" OFF)
+if(NOT SANITIZER)
+  return()
+endif()
+
+#According to: http://clang.llvm.org/docs/AddressSanitizer.html#usage
+# the following compiler flags allow us to keep -O1 on, if desired, but
+# still get pretty call stacks
+set(_SANITIZER_COMPILE_OPTIONS "-g -fno-omit-frame-pointer -fno-optimize-sibling-calls")
+
+string(TOLOWER ${SANITIZER} SANITIZER)
+
+if(CMAKE_COMPILER_IS_GNUCXX_PURE)
+  if(GCC_COMPILER_VERSION VERSION_GREATER 4.7)
+    set(_GCC_SANITIZERS address thread)
+  endif()
+  list(FIND _GCC_SANITIZERS ${SANITIZER} _SANITIZER_FOUND)
+
+elseif(CMAKE_COMPILER_IS_CLANG)
+  if(GCC_COMPILER_VERSION VERSION_GREATER 4.1)
+    set(_CLANG_SANITIZERS address thread undefined)
+  endif()
+  list(FIND _CLANG_SANITIZERS ${SANITIZER} _SANITIZER_FOUND)
+
+endif()
+
+if(${_SANITIZER_FOUND} GREATER -1)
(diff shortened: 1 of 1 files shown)


Sanitizer.cmake
```diff
@@ -0,0 +1,48 @@
...
+endif()
+
+#According to: http://clang.llvm.org/docs/AddressSanitizer.html#usage
+# the following compiler flags allow us to keep -O1 on, if desired, but
+# still get pretty call stacks
+set(_SANITIZER_COMPILE_OPTIONS "-g -fno-omit-frame-pointer -fno-optimize-sibling-calls")
+
+string(TOLOWER ${SANITIZER} SANITIZER)
+
+if(CMAKE_COMPILER_IS_GNUCXX_PURE)
```
eile: missing ```include(Compiler)```

Sanitizer.cmake
```diff
@@ -0,0 +1,48 @@
...
+  list(FIND _CLANG_SANITIZERS ${SANITIZER} _SANITIZER_FOUND)
+
+endif()
+
+if(${_SANITIZER_FOUND} GREATER -1)
+  set(_SANITIZER "-fsanitize=${SANITIZER}")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${_SANITIZER_COMPILE_OPTIONS} ${_SANITIZER}" )
+  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${_SANITIZER}")
+else()
+  message(WARNING "Sanitizer ${SANITIZER} not set, check compiler support")
```
eile: Needed as warning?

Sanitizer.cmake
mgeplf: I'm expecting that to already happen, like in Findcppcheck.cmake and Coverage.cmake, is that not a reasonable assumption?

Sanitizer.cmake
mgeplf: I think so, if you're explicitly setting '-DSANITIZER=foo', it would be nice if I get feedback if that sanitizer doesn't exist for my compiler.  I can drop the warning, if that's better.

Sanitizer.cmake
eile: Depends in which order files are included. Compiler has an early out, so it can't hurt.

Sanitizer.cmake
eile: Ok, fine with me. We might want to enable it later by default, if available.

Sanitizer.cmake
mgeplf: ok, done

Sanitizer.cmake
mgeplf: Turning on the sanitizers comes w/ a performance penalty, so I'm not sure it would be a good idea, even for DEBUG mode.  Doing some performance testing might be worth it, to see if it can be on all the time.

However, having it on for running tests is a good idea, imo.

Sanitizer.cmake
```diff
@@ -0,0 +1,50 @@
...
+#       undefined
+# gcc:
+#    -dumpcompiler: 4.8.3
+#       address
+#       thread
+#
+# Note: This modifies CMAKE_CXX_FLAGS and CMAKE_EXE_LINKER_FLAGS
+# Usage: add this when doing cmake configuration: -DSANITIZER=address
+
+option(SANITIZER "Sanitizer: one of address/thread/undefined, depending on compiler" OFF)
```
eile: ```Provide an option for the user to select as ON or OFF.```

Can't have the described values. What are you trying to achieve? Why not sanitizer on/off, if on add flags?

mgeplf: Yup, my bad, I just dropped the option.

rdumusc:
+1, code is clean. The only limitation that I see it that it only lets you turn on one sanitizer at the time, which wasn't immediately clear to me.

Sanitizer.cmake
rdumusc: I was going to suggest adding back the SANITIZE option as @eile proposed, but since the gcc documentation states "there is no -fsanitize=all, as some sanitizers cannot be used together" I'm sure if this makes sense.

PR #467, opened 2016-03-07
More coverage tweaks

eile:


tribal-tec:
Release notes update? (we need to release CMake as well :p )

eile:
pffuuuohm

eile:
Release note for minor stuff like this?

eile:
Ping?

tribal-tec:
FATAL_ERROR seems worth mentioning, but if it doesn't happen for normal users, +1

eile:
No, doesn't. Only appears on "internal" error but is hard to debug then.

PR #472, opened 2016-04-08
Add common_example

eile:


eile:
Ping - see also 9977f9a

tribal-tec:
Fine with me.
@rdumusc should know for the GUI part I guess

CommonApplication.cmake
```diff
@@ -2,87 +2,89 @@
...
 macro(COMMON_APPLICATION Name)
   common_compiler_flags()
   _common_application(${Name} ${ARGN})
 endmacro()
 
 function(_common_application Name)
+  set(OPT_NAMES GUI EXAMPLE)
+  set(ARG_NAMES)
+  set(ARGS_NAMES)
+  cmake_parse_arguments(THIS "${OPT_NAMES}" "${ARG_NAMES}" "${ARGS_NAMES}"
```
rdumusc: Style:
* remove _NAMES postfix (too many 'name' variables around already)
* use same variable names as in https://cmake.org/cmake/help/v3.2/module/CMakeParseArguments.html ?
* use '_' prefix for local variables

PR #474, opened 2016-04-08
Fixes for install and coverage report in Deflect / Tide

rdumusc:

PR #477, opened 2016-04-13
Rename cpp check targets to fix warning with ITK caused by global namespace pollution

eile:


README.md
```diff
@@ -46,7 +46,7 @@ The following CMake modules can be included in your project:
     * [Coverage](Coverage.cmake) *coverage* target to generate a code coverage
       report as html, if ENABLE_COVERAGE option is also set. Additional compiler
       flags are set in that case, so it should be enabled only for debug builds.
-    * [CppcheckTargets](CppcheckTargets.cmake): *cppcheck* target for
+    * [CommonCPPCheck](CommonCPPCheck.cmake): *cppcheck* target for
       static code analysis. Also adds all cppcheck targets to *tests* target.
     * [clangcheckTargets](clangcheckTargets.cmake): *clangcheck* target for
```
tribal-tec: CommonClangCheck

tribal-tec:
Changelog? Was missing for other commits this week as well I'm afraid (e.g. you +1 a change from @rdumusc)
+1 otherwise

eile:
For this one I figured it's internal, so no ChangeLog needed?

For the other - my bad :-/

rdumusc:
@tribal-tec my bad too; not sure which commit I missed.
@eile I think there are no clear internals in this project and we have seen that some users cherry-pick functions, so it is true that changelog is not a bad idea. (also in this commit it is not clear which was the offending target or function or variable causing global namespace pollution)
Otherwise +1 it is better that way!

eile:
The concrete one was add_cppcheck, ITK has a function of the same name but different impl. Will add changelog.

tribal-tec:
https://github.com/Eyescale/CMake/pull/474 had no changelog

PR #483, opened 2016-04-20
Exclude Objective-C files from cppcheck

tribal-tec:


tribal-tec:
Needed for https://github.com/BlueBrain/Deflect/pull/89

PR #484, opened 2016-04-21
Fix coverage report generation for non-subprojects

tribal-tec:


rdumusc:
This his still had issues, such as:
- did not work if common_library was call from a source folder with a different include path than the bin path (e.g. triply in Equalizer)
- It was stripping the first level of folders potentially causing issues with multiple top-level libraries (e.g. eq + seq)
A solution which (hopefully) addresses theses shortcomings can be found in #486, however this comes at the cost of enforcing that the subprojects must be placed outside the project source tree (using a superproject or the COMMON_SOURCE_DIR to place them outside). This seems unfortunately unavoidable.

PR #485, opened 2016-04-22
Fix Doxygen's project version after CMake3 port

rdumusc:


DoxygenRule.cmake
```diff
@@ -77,6 +81,10 @@ if(NOT DOXYGEN_PROJECT_NAME)
   set(DOXYGEN_PROJECT_NAME ${PROJECT_NAME})
 endif()
 
+if(NOT DOXYGEN_PROJECT_VERSION)
+  set(DOXYGEN_PROJECT_VERSION ${${PROJECT_NAME}_VERSION})
```
hernando: Why not ${PROJECT_VERSION}?

hernando:
Apart from the minor comment +1

DoxygenRule.cmake
eile: CMake 3 has the version in the project() and exports it into Project_VERSION

PR #486, opened 2016-04-22
Fix coverage report generation for the top-level project

rdumusc:


CHANGES.md
```diff
@@ -1,5 +1,7 @@
 # git master
 
+* [486](https://github.com/Eyescale/CMake/pull/486):
+  Fix coverage report generation for non-subprojects
```
hernando: Indeed, you have solved it for the subproject case. Don't you mean "for non-superprojects"?

rdumusc: This is from Daniel's commit but it is correct. It was already working for the subproject case (i.e. a project which is a subproject of viz/stable for instance). Now it is fixed (with a small limitation) for projects which are not a subproject but are their own superproject (thus non-subprojects). Nomenclature is a bit tricky for sure :-)

hernando: We have to agree on the nomenclature, because what you call the subproject case (viz/stable) for me it's the superproject case.

rdumusc: Or: fix coverage report generation for the top-level project?

hernando: I prefer that one.

rdumusc:
updated

hernando:
+1

PR #488, opened 2016-04-29
Fix usage of VERSION -> PROJECT_VERSION after CMake3 change

rdumusc:
CommonApplication part was needed to deploy DesktopStreamer, but I updated all the remaining occurrences...


jafyvilla:
👍

PR #491, opened 2016-05-19
Fix regression, fix multiple install of defines.h

tribal-tec:


rdumusc:
+1 for the fix itself, but shouldn't we just remove this "per-system" defines file and keep just defines.h?

tribal-tec:
Not sure as well. I kept it defensive and maintained what was used before. I would postpone this potential cleanup and add an issue to resolve this. There might be things that I/we miss now, but @eile might know.

PR #493, opened 2016-05-25
Do not mess with LCOV_EXCLUDE which is set by outside users

tribal-tec:
After #486, excluding the tests is not needed anymore.


rdumusc:
+1 good catch one more ugly hack which can go away!

PR #498, opened 2016-06-28
Add Findlibmemcached.cmake

tribal-tec:


eile:
Oh god, really? My installation has ```/opt/local/lib/pkgconfig/libmemcached.pc``` and works without a finder - should we rather not create this one in the installation?

Otherwise +1

tribal-tec:
Oh, good catch, haven't seen it.

PR #500, opened 2016-07-05
Added WIN32 option to CommonApplication to be able to build WinMain-b…

ptoharia:
…ased apps (no console).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,8 @@
 # 2016.06 (30-Jun-2016)
 
+* [500](https://github.com/Eyescale/CMake/pull/500):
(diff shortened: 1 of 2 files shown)


rdumusc:
Strange, this WIN32 flag was always there for GUI applications. After searching I discovered that it was removed here: https://github.com/Eyescale/CMake/pull/480

tribal-tec:
Before WIN32 was always given to add_executable which didn't work if the app was already using Qt. If specified from outside as an option to common_gui_application, it's fine I guess.

tribal-tec:
Please adapt the CHANGES.md accordingly.

ptoharia:
@tribal-tec done

PR #501, opened 2016-07-20
Add FindCairo.cmake and poppler-glib component to FindPoppler.cmake

rdumusc:
I notice now that there was already a Cairo finder, but it was not working for me, so I used that one.


tribal-tec:
pkg_check_modules(cairo) is not sufficient or why is the finder needed?

rdumusc:
Yes, the finder are not needed! I checked on the cluster too, I really wonder why we spent so much time writing those... see https://github.com/Eyescale/CMake/pull/502

PR #503, opened 2016-07-25
Add MODULE option to common_find_package, simplify argument parsing

rdumusc:
Seems to work but please review carefully :-)

### DIFF
--- a/CommonFindPackage.cmake
+++ b/CommonFindPackage.cmake
@@ -45,54 +45,46 @@ option(COMMON_FIND_PACKAGE_QUIET "Use QUIET for common_find_package command" ON)
 
 macro(common_find_package Package_Name)
   string(TOUPPER ${Package_Name} PACKAGE_NAME)
-  set(__args ${ARGN}) # ARGN is not a list. make one.
 
-  # SYSTEM specified?
-  set(__is_system_package)
-  if(__args)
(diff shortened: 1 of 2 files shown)


CommonFindPackage.cmake
```diff
@@ -45,54 +45,46 @@ option(COMMON_FIND_PACKAGE_QUIET "Use QUIET for common_find_package command" ON)
...
+  # Parse macro arguments
+  set(__options QUIET REQUIRED SYSTEM)
+  set(__oneValueArgs MODULE)
+  set(__multiValueArgs)
+  cmake_parse_arguments(__pkg "${__options}" "${__oneValueArgs}" "${__multiValueArgs}" ${ARGN})
 
   # Boost is always SYSTEM
   if(${Package_Name} STREQUAL "Boost")
-    set(__is_system_package ON)
+    set(__pkg_SYSTEM TRUE)
```
tribal-tec: __pkg_SYSTEM is not used. There must be a __is_system_package at the end of the function which does include_directories(BEFORE SYSTEM ${__include_dirs}

tribal-tec:
+1
I hope I was carefully enough

CommonFindPackage.cmake
rdumusc: Good catch!

PR #504, opened 2016-08-03
Improvements in the package installation feature.

hernando:
Support for yum and avoid caching of INSTALL_PACKAGES in CMakeCache.txt.
This changes are motivated by some issues reported by a user trying
to compile Equalizer in RHEL.

### DIFF
--- a/SubProject.cmake
+++ b/SubProject.cmake
@@ -37,8 +37,11 @@
 # - DISABLE_SUBPROJECTS: when set, does not load sub projects. Useful for
 #   example for continuous integration builds
 # - SUBPROJECT_${name}: If set to OFF, the subproject is not added.
-# - INSTALL_PACKAGES: command line cache variable which will "apt-get" or
-#   "port install" the known system packages. Will be unset after installation.
+# - INSTALL_PACKAGES: command line cache variable which will "apt-get", "yum" or
+#   "port install" the known system packages. This variable is unset after this
+#   script is parsed by top level projects.
+#   The packages to install are taken from ${PROJECT_NAME}_${type}_DEPENDS
+#   where type is DEB, RPM or PORT depending on the system.
 # - COMMON_SOURCE_DIR: When set, the source code of subprojects will be
 #   downloaded in this path instead of CMAKE_SOURCE_DIR.
 # A sample project can be found at https://github.com/Eyescale/Collage.git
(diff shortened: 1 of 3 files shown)


hernando:
Wait a moment. This code is leaving __pkg_mng set and it appears in the ccmake menu.

SubProject.cmake
```diff
@@ -80,12 +83,33 @@ function(subproject_install_packages name)
...
-    list(SORT ${NAME}_DEB_DEPENDS)
-    list(REMOVE_DUPLICATES ${NAME}_DEB_DEPENDS)
-    message("Running 'sudo apt-get install ${${NAME}_DEB_DEPENDS}'")
-    execute_process(COMMAND sudo apt-get install ${${NAME}_DEB_DEPENDS})
+  if(CMAKE_SYSTEM_NAME MATCHES "Linux" )
+    # Detecting the package manager to use
+    find_program(__pkg_mng apt-get)
+    if(__pkg_mng)
+      set(__pkg_type DEB)
+    elseif(__pkg_mng)
```
rdumusc: else()

SubProject.cmake
```diff
@@ -80,12 +83,37 @@ function(subproject_install_packages name)
...
-    list(SORT ${NAME}_DEB_DEPENDS)
-    list(REMOVE_DUPLICATES ${NAME}_DEB_DEPENDS)
-    message("Running 'sudo apt-get install ${${NAME}_DEB_DEPENDS}'")
-    execute_process(COMMAND sudo apt-get install ${${NAME}_DEB_DEPENDS})
+  if(CMAKE_SYSTEM_NAME MATCHES "Linux" )
+    # Detecting the package manager to use
+    find_program(__pkg_mng apt-get)
+    if(__pkg_mng)
+      set(__pkg_type DEB)
+    else(__pkg_mng)
```
rdumusc: please don't repeat the condition in the else()

SubProject.cmake
```diff
@@ -80,12 +83,37 @@ function(subproject_install_packages name)
...
+      find_program(__pkg_mng yum)
+      if(__pkg_mng)
+        set(__pkg_type RPM)
+      endif()
+    endif()
+    if(NOT __pkg_mng)
+      message(WARNING "Could not determine the package manager for installing dependencies")
```
rdumusc: could we print the warning on APPLE as well if macports is not found?

CommonFindPackage.cmake
```diff
@@ -269,4 +269,8 @@ macro(common_find_package_post)
   endif()
   common_graph(${PROJECT_NAME})
   message(STATUS ${__configure_msg})
+
+  # If this variable was given in the command line, ensure that the package
+  # installation is only run in this cmake invocation.
+  unset(INSTALL_PACKAGES CACHE)
```
rdumusc: Is there no way to move this to SubProject.cmake ? Some users of the Subproject feature may not use CommonFindPackage (or not correctly)

rdumusc:
looks good otherwise

CommonFindPackage.cmake
hernando: I don't think this can be moved to SubProject.cmake because it has to be run just once and after all subprojects have been processed. This seems to be the only place where this can be guaranteed.
I thought about it, but if you are not using CommonFindPackage correctly this is the smallest of your problems. On the other hand, does it make sense to use this project if you're not going to use common_find_package? If you are cherry picking files you can take care of this details yourself as well.

rdumusc: Limiting interactions across the different files really helps with modularity and maintainability.
After looking a bit I think that it can in fact be moved to the end of SubProject.cmake. We could make a check for ${name}_IS_SUBPROJECT to guarantee that it is only executed by the top-level project.

rdumusc: try adding the following at the end of SubProject.cmake and see the output of the configure step:

if(NOT DEFINED ${PROJECT_NAME}_IS_SUBPROJECT)
    message(WARNING "SubProject.cmake done: ${PROJECT_NAME} - subproject: ${${PROJECT_NAME}_IS_SUBPROJECT}")
endif()

SubProject.cmake
```diff
@@ -37,8 +37,11 @@
 # - DISABLE_SUBPROJECTS: when set, does not load sub projects. Useful for
 #   example for continuous integration builds
 # - SUBPROJECT_${name}: If set to OFF, the subproject is not added.
-# - INSTALL_PACKAGES: command line cache variable which will "apt-get" or
-#   "port install" the known system packages. Will be unset after installation.
+# - INSTALL_PACKAGES: command line cache variable which will "apt-get", "yum" or
+#   "port install" the known system packages. This variable is unset from
+#   common_find_package_post.
```
rdumusc: missing doc update

CommonFindPackage.cmake
```diff
@@ -269,4 +269,5 @@ macro(common_find_package_post)
   endif()
   common_graph(${PROJECT_NAME})
   message(STATUS ${__configure_msg})
+
```
rdumusc: no changes in this file

SubProject.cmake
```diff
@@ -241,4 +272,11 @@ if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/.gitsubprojects")
     endif()
     add_dependencies(update ${PROJECT_NAME}-update-git-subprojects)
   endif()
+
+  if(NOT ${name}_IS_SUBPROJECT)
```
rdumusc: Should be ${PROJECT_NAME}_IS_SUBPROJECT. Otherwise you actually rely on ${name} being undefined...

PR #505, opened 2016-08-03
Some common tricks needed for projects using CUDA.

hernando:
The tricks are necessary to avoid warnings in some cases.


hernando:
Please review @rdumusc

rdumusc:
+1, maybe just check if the search for CUDA_HOST_COMPILER does not happen for each subproject after CUDA_FOUND is true.

PR #507, opened 2016-08-30
Handle required version in FindNumPy.cmake

hernando:


tribal-tec:
Changelog update missing. Also mention there the change of the NUMPY_FOUND variable.

hernando:
I haven't really changed anything regarding NUMPY_FOUND. The variable is still set, but the script was setting PYTHON_NUMPY_FOUND manually also. I think no project is using the latter, but I decided not to remove it just in case and I updated the documentation of the script instead.

tribal-tec:
+1

hernando:
Please merge, I can't do it.

PR #508, opened 2016-09-01
Remove eof at file path to feed in find_path hints.

hernando:


FindNumPy.cmake
```diff
@@ -18,19 +18,19 @@ if (PYTHON_EXECUTABLE)
...
-            "try: import numpy; print(numpy.__version__)\nexcept:pass\n"
+            "from __future__ import print_function\ntry: import numpy; print(numpy.__version__, end='')\nexcept:pass\n"
     OUTPUT_VARIABLE __numpy_version)
 elseif(__numpy_out)
   message(STATUS "Python executable not found.")
 endif(PYTHON_EXECUTABLE)
 
 find_path(PYTHON_NUMPY_INCLUDE_DIR numpy/arrayobject.h
-  HINTS "${__numpy_path}" "${PYTHON_INCLUDE_PATH}")
+  HINTS "${__numpy_path}" "${PYTHON_INCLUDE_PATH}" NO_DEFAULT_PATH)
```
tribal-tec: NO_DEFAULT_PATH should be removed, no?

hernando: Not sure... if find_path fails (which shouldn't) what behaviour do you prefer, that the include path is missing or that it may point to an incorrect location?

PR #510, opened 2016-09-22
Spec for sub target modularization

eile:
pretty-formatted: https://github.com/eile/CMake/blob/master/doc/modules.md


doc/modules.md
```diff
@@ -0,0 +1,88 @@
...
+    Lunchbox::PersistentMap
+    Lunchbox::Plugin
+    Lunchbox::Process (daemon+launcher)
+    Lunchbox::RNG
+    Lunchbox::Servus
+    Lunchbox::TLS
+    Lunchbox::Test
+    Lunchbox::Thread
+    Lunchbox::URI
+    Lunchbox::UnorderedIntervalSet
```
hernando: After checking the code with Daniel I think this needs be renamed to OrderedIntervalSet; there was an error when the class was moved from the original project into Lunchbox.

doc/modules.md
```diff
@@ -0,0 +1,88 @@
...
+  lunchbox/log/log.h and .cpp. The new common_add_library() cmake module
+  auto-generates the top-level header. New sub libraries set
+  NAME_OMIT_LIBRARY_HEADER.
+* The old common_library is deprecated, and stays during the transition
+  period.
+* Optional sub-targets are to be avoided and hidden by facades (esp. in
+  Lunchbox, see PersistentMap). If they are necessary, e.g, in hwsd,
+  downstream code needs to test for the target's availability (as done
+  e.g. in eq/server).
+* Servus will eventually be reintegrated as a lunchbox sub library.
```
rdumusc: Freedom for Servus!! 😮

doc/modules.md
```diff
@@ -0,0 +1,88 @@
...
+    Lunchbox::UnorderedIntervalSet
+
+## Issues
+
+### 1: How do handle optional targets?
+
+_Resolution: Avoid them as possible, downstream projects need to check
+with if(TARGET) in cmake:_
+
+We could (re)use the COMPONENTS parameter in common_find_package(), but
```
rdumusc: I would rather suggest to do like Qt and create one find script per target, which avoids the problem and allows much finer grained use in downstream projects. Real world example:

common_find_package(Qt5Core REQUIRED)
common_find_package(Qt5WebEngine 5.6)
if(NOT TARGET Qt5::WebEngine)
  common_find_package(Qt5WebKitWidgets SYSTEM)
endif()

doc/modules.md
```diff
@@ -0,0 +1,88 @@
...
+* All sub project targets are excluded from the default build, and
+  required targets are built only based on the dependencies.
+* Sub-libraries have a top-level header, e.g., lunchbox/log.h, and a
+  sub-directory with the implementation and CMakeLists, e.g.,
+  lunchbox/log/log.h and .cpp. The new common_add_library() cmake module
+  auto-generates the top-level header. New sub libraries set
+  NAME_OMIT_LIBRARY_HEADER.
+* The old common_library is deprecated, and stays during the transition
+  period.
+* Optional sub-targets are to be avoided and hidden by facades (esp. in
```
rdumusc: Doesn't sound very practical, see my comment below about having one find script per library instead.

doc/modules.md
```diff
@@ -0,0 +1,88 @@
...
+    common_add_library(Lunchbox DEPENDS PUBLIC Lunchbox::Log)
+
+Generates: .../include/lunchbox.h
+
+## Examples
+
+    #include <zeroeq.h>
+    #include <lunchbox/log.h>
+
+### Desired Lunchbox Granularity
```
rdumusc: Looks great!

doc/modules.md
```diff
@@ -0,0 +1,88 @@
...
+
+    common_add_library(Lunchbox::Log HEADERS logImpl.h PUBLIC_HEADERS log.h
+      SOURCES log.cpp DEPENDS PUBLIC Lunchbox::Bar PRIVATE Lunchbox::Foo)
+
+Generates: .../include/lunchbox/log.h, .../include/lunchbox/log/log.h,
+.../lib/libLunchboxLog.so
+
+    common_add_library(Lunchbox DEPENDS PUBLIC Lunchbox::Log)
+
+Generates: .../include/lunchbox.h
```
rdumusc: I wonder if this one is actually useful

doc/modules.md
eile: Added

doc/modules.md
eile: We discussed this, see added comment in Issue 1. We already did it like this for other reasons, but let's discuss on Monday.

doc/modules.md
eile: Freedom is Slavery!! We've always been happy with Lunchbox::Servus!!

(not sure how to read your comment, so going full Orwell here)

doc/modules.md
eile: We use it quite a lot, if you just don't want to care about sub components.

doc/modules.md
eile: See above and updated spec.

eile:
@favreau FYI

doc/modules.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Requirements
+
+* Per-functionality targets
+* Improved build times for super projects
+
+## Implementation
+
+* All sub project targets are excluded from the default build, and
+  required targets are built only based on the dependencies.
```
hernando: What if you are building Lunchbox as a standalone project to install the libraries in your system and use them from there afterwards?

doc/modules.md
```diff
@@ -0,0 +1,96 @@
...
+* Optional sub-targets are to be avoided and hidden by facades (esp. in
+  Lunchbox, see PersistentMap). If they are necessary, e.g, in hwsd,
+  downstream code needs to test for the target's availability (as done
+  e.g. in eq/server).
+* Servus will eventually be reintegrated as a lunchbox sub library.
+
+## Naming
+
+    common_add_library(Lunchbox::Log HEADERS logImpl.h PUBLIC_HEADERS log.h
+      SOURCES log.cpp DEPENDS PUBLIC Lunchbox::Bar PRIVATE Lunchbox::Foo)
```
hernando: If there any magic associated with the :: or is it just a convention because CMake doesn't tokenize target identifiers at :?

doc/modules.md
```diff
@@ -0,0 +1,96 @@
...
+few use cases (hwsd, ZeroEQ Qt support). We might decide to improve
+common_find_package() later if needed.
+
+Alternatively we could generate a package config for each sub library,
+so that downstream projects check for each sub library individually,
+e.g., common_find_package(Lunchbox::Log REQUIRED). In consequence this
+means that each sub library becomes a separate git project: find_package
+is designed to find a project, and we eventually want to extend the
+common_find_package syntax to include a git url and SHA to get rid of
+the redundant .gitsubprojects.
```
hernando: I guess that what a package is is ill-defined in general, so we should agree on what does it really mean. Qt is not the only example where there's no one-to-one relationship between repository - git module - CMake package, VTK is similar in this regard.

Whatever is implemented, I'd prefer to have a single place where the required Lunchbox subtargets are requested. It can be done just after common_find_package_post, but is looks more correct to do what Raphael suggests or use COMPONENTS (which by the way I don't understand why you say it overloads semantics, Boost works exactly that way).

doc/modules.md
```diff
@@ -0,0 +1,96 @@
...
+    Lunchbox::PersistentMap
+    Lunchbox::Plugin
+    Lunchbox::Process (daemon+launcher)
+    Lunchbox::RNG
+    Lunchbox::Servus
+    Lunchbox::TLS
+    Lunchbox::Test
+    Lunchbox::Thread
+    Lunchbox::URI
+    Lunchbox::OrderedIntervalSet (was UnorderedIntervalSet)
```
hernando: Do you have any idea of whether this level of fragmentation can affect the loading time of applications in a noticeable way?

doc/modules.md
eile: Then Lunchbox is the superproject and all targets are in the default build.

doc/modules.md
eile: Iirc it's the CMake convention/namespacing for sub-targets. We found this somewhere in the CMake documentation dungeons. @tribal-tec: do you remember there?

doc/modules.md
eile: I think it's mostly a non-issue today. The only optional targets I am aware of is hwsd and ZeroEQ::Qt support today, and they are already handled downstream correctly with nobody complaining.

hernando: Then, I understood it wrong. When Lunchbox is detected with common_find_package it won't be necessary to test for Lunchbox::Foo because everything is going to be there, right? Just that it won't be built by default if not needed.

doc/modules.md
tribal-tec: Since CMake 3, it denotes imported or alias targets: https://cmake.org/cmake/help/v3.0/policy/CMP0028.html. We will/have to use for export() of each subtarget.

doc/modules.md
tribal-tec: Or just IntervalSet. std.set is also ordered, unordered_set not.

doc/modules.md
tribal-tec: On slow filesystems potentially. If this really is an issue, static linking is your friend :imp:

doc/modules.md
rdumusc: I haven't given much thought about how the merging of common_find_package and .gitsubproject will look like in practice. Maybe it should be a new command, that calls  common_find_package internally to avoid the repetition but does not cause confusion with finding system packages. In any case, I don't think it's a blocker, and that it implies creating separate git projects. I does point to something interesting though, that maybe each library could be versionned separately.
I agree with Juan that what a package is is ill-defined in general. For instance, boost or qt come as several pacakges on linux, but as one install on macports. Also Poppler, which can be considered as one small project, but is split into Poppler, poppler-qt, poppler-glib, etc...
COMPONENTS finding look like the right thing to do at first but they are a pain to handle in practice. I know it from writing a FindPoppler.cmake with separate components and from the problems that we have (had) with finding all those boost components. The way Qt works now is much better IMO.
On the contrary, there are plenty of such optional targets: ServusQt, DeflectServer (today part of Deflect, but should be separated), probably 90% of the upcoming Lunchbox targets, etc.

doc/modules.md
rdumusc: I would argue that generally you should care about including only the components you use. Could be OK for a small project, but I wouldn't include boost/boost.h for instance. But that's just a detail.

doc/modules.md
rdumusc: There is no magic to ::, but yes I think it's a convention. find_package(Qt5Core) -> Qt5::Core.

rdumusc: thanks @tribal-tec for the clarification, althogh I am not sure I understand the docuementation correcly. So inside the Lunchbox project, are the targets named differently (wihtout the ::) because it is not an imported target? So there is LunchboxLog depends on LunchboxAny instead of Lunchbox::Any, for instance?

doc/modules.md
eile: Yes, there is a static per-shared-library ld.so cost. Shall we use static linking? ;)

doc/modules.md
eile: There is no test for Lunchbox::Foo. Config mode package configs simply declare an imported target Lunchbox::Foo.

doc/modules.md
eile: Done

CommonCompiler.cmake
```diff
@@ -1,12 +1,13 @@
...
 #
 # Sets compiler optimization, definition and warnings according to
 # chosen compiler. Supported compilers are XL, Intel, Clang, gcc (4.4
 # or later) and Visual Studio (2008 or later).
 #
-# This defines the common_compile_options() function to apply compiler flags and
-# features for the given target.
+# This defines the common_compile_options() function to apply compiler
+# flags and features for the given target, and adds the target to the
+# project-all meta target.
```
rdumusc: I don't think adding the target to the project-all target has anything to do with the compile options (setting compilation flags).
I would move this to the three places where this function is needed: CommonApplication, CommonLibrary, and CommonCPPCtest, probably where  ${PROJECT_NAME}_ALL_DEP_TARGETS is already filled.

rdumusc: In fact, couldn't the ${PROJECT_NAME}_ALL_DEP_TARGETS be removed entirely? DoxygenRule could also depend on ${PROJECT_NAME}-all, right?

eile: Ok, I placed it in compile_flags to not C&P it in three places, but since you are the second one to complain I moved it up now. ;)

Common.cmake
```diff
@@ -115,6 +117,12 @@ if(NOT PROJECT_namespace)
   set(PROJECT_namespace ${PROJECT_INCLUDE_NAME})
 endif()
 
+if(NOT TARGET ${PROJECT_NAME}-all)
+  # Create <project>-all target. Deps are added by common_compile_options()
```
rdumusc: not anymore

CommonCompiler.cmake
rdumusc: great! then this file should be reverted (no changes, comment no longer applies)

DoxygenRule.cmake
```diff
@@ -103,35 +98,21 @@ if(NOT DOXYGEN_CONFIG_FILE)
...
 
-if(NOT TARGET doxygen_html)
-  add_custom_target(doxygen_html)
-  set_target_properties(doxygen_html PROPERTIES
-    EXCLUDE_FROM_DEFAULT_BUILD ON FOLDER ${PROJECT_NAME}/doxygen)
-endif()
-add_dependencies(doxygen_html doxygen_html_${PROJECT_NAME})
-
-add_custom_target(${PROJECT_NAME}-doxygen DEPENDS doxygen_html_${PROJECT_NAME})
+add_custom_target(${PROJECT_NAME}-doxygen DEPENDS ${PROJECT_NAME}-html)
```
rdumusc: This looks much better! Now do we need to keep two separate targets which do the same thing? I would keep only ${PROJECT_NAME}-doxygen and remove ${PROJECT_NAME}-html.
Also, the new PROJECT-install target is missing from the documentation above.

eile: The -html and -install targets are internal and undocumented on purpose.

Removed -html target.

PR #516, opened 2016-10-28
Support for GCC 6

eile:


hernando:
This is a bit weird, I've been able to compile many of our tools with gcc 6.0 without making any changes. Which is the system in which you're having problems.

eile:
Hold this one. Won't work on Ubuntu

eile:
More better. Proceed at will.

CommonCompiler.cmake
```diff
@@ -41,6 +41,11 @@ endif()
 option(COMMON_WARN_DEPRECATED "Enable compiler deprecation warnings" ON)
 option(COMMON_ENABLE_CXX11_STDLIB "Enable C++11 stdlib" OFF)
 option(COMMON_DISABLE_WERROR "Disable -Werror flag" OFF)
+if($ENV{CMAKE_COMMON_USE_CXX03_ABI})
```
eile: erm yes. Set by env module.

PR #518, opened 2016-11-03
Use common_find_package for Doxygen

hernando:


tribal-tec:
Why needed?

hernando:
Because in Brion we have: 

    #if defined BRION_USE_SPHINX && defined BRION_USE_DOXYGEN

And BRION_USE_DOGYXGEN doesn't get defined anywhere (actually it gets, but in a nested scoped). This is the simplest solution and with a previous commit that reduces the verbosity of the configuration, you won't notice the difference.

hernando:
The code inside the if is used to remove the C++ signatures from the Python wrapping docs. Look [here](https://developer.humanbrainproject.eu/docs/projects/Brion/1.9/python/index.html) to understand why I want to get rid of them.

tribal-tec:
A speciality in one project is not common to me. If there are no side-effects of using common_find_package here, fine with me.

hernando:
I can find an alternative, but it will look uglier in Brion.

rdumusc:
I don't know if I'm missing something, but this looks wrong. It will cause all the projects to print "configured with/without Doxyen". I don't even understand why there is this find_package(Doxygen QUIET) in Common.cmake because I can't see it being used anywhere outside Doxygen.cmake, where there is another find_package(Doxygen) anyway.
I would:
- remove this entry from Common.cmake, or if not possible document/justify why it's there
- do anything specific to Brion in Brion - for instance a  common_find_package(Doxygen QUIET) in the main CMakeLists.txt or setting BRION_USE_DOXYGEN as a compile definition if DOXYGEN_FOUND

tribal-tec:
cf first bullet-point:
The find_package(Doxygen) is top-level to do it only for a super-project setup and not every time in DoxyGen.cmake which are in the projects and hence out-of-scope for other projects. It's an optimization to reduce the CMake runtime after I did some profiling/benchmarking in the past.

PR #523, opened 2016-11-09
Add find_cuda_device()

jafyvilla:
Compile and run a simple CUDA application to detect whether
a CUDA-capable device is present or not.

### DIFF
--- a/CommonCUDA.cmake
+++ b/CommonCUDA.cmake
@@ -1,12 +1,16 @@
 # Copyright (c) 2016 Juan.Hernando@epfl.ch
-
-# Provides two functions for common configuration checks and setup for projects
-# using CUDA.
+#               2016 Jafet.VillafrancaDiaz@epfl.ch
+#
+# Provides three functions for common configuration checks and setup for
+# projects using CUDA.
 # In particular:
 # * find_cuda_compatible_host_compiler() tries to set CUDA_HOST_COMPILER to a
 #   version compatible with the CUDA version detected.
 # * common_cuda_compile_options() sets the default architecture to a minimum
 #   value that does not cause deprecation warnings with newer version of nvcc.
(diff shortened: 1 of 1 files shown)


CommonCUDA.cmake
```diff
@@ -100,4 +104,26 @@ function(common_cuda_compile_options)
...
+
+  try_run(RUN_RESULT_VAR COMPILE_RESULT_VAR
+          ${CMAKE_CURRENT_BINARY_DIR}/hasCudaDevice
+          SOURCES ${CMAKE_CURRENT_BINARY_DIR}/hasCudaDevice.cpp
+          CMAKE_FLAGS -DINCLUDE_DIRECTORIES=${CUDA_INCLUDE_DIRS}
+                      -DLINK_LIBRARIES=${CUDA_LIBRARIES}
+          COMPILE_OUTPUT_VARIABLE COMPILE_OUTPUT_VAR
+          RUN_OUTPUT_VARIABLE RUN_OUTPUT_VAR)
+  if(COMPILE_RESULT_VAR AND NOT RUN_RESULT_VAR)
+    set(CUDA_DEVICE_FOUND TRUE CACHE BOOL "Whether a CUDA-capable device is present")
```
eile: Do you need this to be a cache variable? They always are troublesome for me.

jafyvilla: I thought it could be convenient if several projects run it in a same configuration step (stable?). I don't really need it, you are the CMake experts, so I'll change it if you think it hurts more than anything.

eile: You still run the test multiple times. For now, just use ```set(CUDA_DEVICE_FOUND TRUE|FALSE)```. We can optimize later if need be.

jafyvilla: done

PR #526, opened 2016-11-28
Provide getSchema() and toJSON() for generated version.h to ease usage in HTTP server

tribal-tec:


tribal-tec:
Gives linker errors as the version file is rather messy regarding sublibraries. So no, not moving to cpp.

PR #528, opened 2016-12-05
Release preparation

tribal-tec:


CommonCompiler.cmake
```diff
@@ -8,6 +8,13 @@
 # This defines the common_compile_options() function to apply compiler flags and
 # features for the given target.
 #
+# CMake options:
+# * COMMON_WARN_DEPRECATED: Enable compiler deprecation warnings, default ON
+# * COMMON_ENABLE_CXX11_STDLIB: Enable C++11 stdlib, default OFF
```
hernando: I don't understand why this is OFF by default.

rdumusc: I'm not entirely sure, but I think it is to ensure interoperability with system packages which use the old stdlib

PR #532, opened 2017-01-11
Fix missing submodule update after rebase, allow more customized .gitsubprojects

tribal-tec:


GitExternal.cmake
```diff
@@ -112,6 +112,9 @@ function(GIT_EXTERNAL DIR REPO tag)
       RESULT_VARIABLE nok ERROR_VARIABLE error WORKING_DIRECTORY "${DIR}")
     if(nok)
       message(FATAL_ERROR "git checkout ${TAG} in ${DIR} failed: ${error}\n")
+    else()
+      execute_process(COMMAND "${GIT_EXECUTABLE}" submodule update --init
```
eile: add --recursive

GitExternal.cmake
```diff
@@ -180,8 +183,7 @@ function(GIT_EXTERNAL DIR REPO tag)
       "  RESULT_VARIABLE nok ERROR_VARIABLE error\n"
       "  WORKING_DIRECTORY \"${DIR}\")\n"
       "if(nok)\n"
-      "  message(FATAL_ERROR \"git checkout ${TAG} in ${__dir} failed: ${error}\n\")\n"
-      "endif()\n"
+      "  message(FATAL_ERROR \"git checkout ${TAG} in ${__dir} failed: \${error}\")\n"
     )
```
tribal-tec: No, continues with else()endif() in l.208

GitExternal.cmake
tribal-tec: Done

GitExternal.cmake
eile: Yes, reviewing offline now. Much easier to see.

SubProject.cmake
```diff
@@ -51,12 +51,10 @@
...
 include(${CMAKE_CURRENT_LIST_DIR}/CommonGraph.cmake)
 
-if(TARGET git_subproject_${PROJECT_NAME}_done)
+if(GIT_SUBPROJECT_DONE_${PROJECT_NAME})
   return()
 endif()
-add_custom_target(git_subproject_${PROJECT_NAME}_done)
-set_target_properties(git_subproject_${PROJECT_NAME}_done PROPERTIES
-  EXCLUDE_FROM_DEFAULT_BUILD ON FOLDER ${PROJECT_NAME}/zzphony)
+set(GIT_SUBPROJECT_DONE_${PROJECT_NAME} ON)
```
eile: That does not have global scope as the target does.

tribal-tec: target is still the wrong thing here. let's see offline

GitExternal.cmake
```diff
@@ -112,6 +112,9 @@ function(GIT_EXTERNAL DIR REPO tag)
       RESULT_VARIABLE nok ERROR_VARIABLE error WORKING_DIRECTORY "${DIR}")
     if(nok)
       message(FATAL_ERROR "git checkout ${TAG} in ${DIR} failed: ${error}\n")
+    else()
+      execute_process(COMMAND "${GIT_EXECUTABLE}" submodule update --init --recursive
+        WORKING_DIRECTORY "${DIR}")
```
hernando: Do you think this solves the arbitrary failures with parallel clones?

GitExternal.cmake
```diff
@@ -112,6 +112,9 @@ function(GIT_EXTERNAL DIR REPO tag)
       RESULT_VARIABLE nok ERROR_VARIABLE error WORKING_DIRECTORY "${DIR}")
     if(nok)
       message(FATAL_ERROR "git checkout ${TAG} in ${DIR} failed: ${error}\n")
```
hernando: return after fatal errors?
Why in line 103 the message is not a fatal error?

SubProject.cmake
```diff
@@ -157,11 +155,12 @@ if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/.gitsubprojects")
     set(GIT_SUBPROJECTS_SCRIPT
       "${CMAKE_CURRENT_BINARY_DIR}/UpdateSubprojects.cmake")
     file(WRITE "${GIT_SUBPROJECTS_SCRIPT}"
-      "file(WRITE .gitsubprojects \"# -*- mode: cmake -*-\n\")\n")
+      "file(READ .gitsubprojects GIT_SUBPROJECTS_FILE)\n")
```
hernando: What if the file doesn't exist at all (e.g. a project without subprojects)?

tribal-tec: l.151: if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/.gitsubprojects")

GitExternal.cmake
tribal-tec: Maybe. You can retry if you want :)

GitExternal.cmake
tribal-tec: in 103 it's not an error for optional externals. in 106 there's a fatal error. In doubt ask the original author :)

SubProject.cmake
tribal-tec: Done

SubProject.cmake
hernando: That was not visible in the diff :)

GitExternal.cmake
hernando: If I don't have anything more important to do I'll try :P

PR #535, opened 2017-01-23
Extracted common_help() out of common_application()

rdumusc:
This allows the generation of help pages for applications not compiled with common_application, such as python scripts (tide, rtneuron-app.py).

### DIFF
--- a/CommonHelp.cmake
+++ b/CommonHelp.cmake
@@ -1,0 +1,77 @@
+# Copyright (c) 2017 Stefan.Eilemann@epfl.ch
+#                    Raphael.Dumusc@epfl.ch
+#
+# Generate help page for doxygen by running Name application with --help:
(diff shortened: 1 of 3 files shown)


CommonHelp.cmake
```diff
@@ -0,0 +1,76 @@
+# Copyright (c) 2017 Stefan.Eilemann@epfl.ch
+#                    Raphael.Dumusc@epfl.ch
+#
+# Generate help page for doxygen by running Name application with --help:
```
eile: Document that Name is the target name

rdumusc: Yes, this was not clear. I added to the Arguments list:
   Name: an existing application target

eile: Ah, below. nevermind.

PR #538, opened 2017-01-30
Allow integration of subprojects outside of COMMON_SOURCE_DIR

tribal-tec:


tribal-tec:
Needed in https://github.com/HBPVIS/ZeroEQ/pull/198/files

hernando:
Late to the party, but doesn't any documentation need to be updated after this change?

tribal-tec:
add_subproject is a semi-public function and nothing is mentioned about the path in its 'documentation'. So, no regression, but also no improvement :)

PR #544, opened 2017-03-15
FindLibJpegTurbo.cmake parses jconfig.h for version check

rdumusc:
This is required because LibJpegTurbo's pkgconfig prior to version 1.5 (e.g. 1.4.2 on Ubuntu 16.04) is missing version information. Earlier versions may not even have a pkgconfig at all. This change enables users to require a minimum version with find_package(LibJpegTurbo).

### DIFF
--- a/FindLibJpegTurbo.cmake
+++ b/FindLibJpegTurbo.cmake
@@ -4,6 +4,7 @@
 #  LibJpegTurbo_FOUND - system has libjpeg-turbo
 #  LibJpegTurbo_INCLUDE_DIRS - the libjpeg-turbo include directories
 #  LibJpegTurbo_LIBRARIES - link these to use libjpeg-turbo
+#  LibJpegTurbo_VERSION - the version of libjpeg-turbo
 #
 # this file is modeled after http://www.cmake.org/Wiki/CMake:How_To_Find_Libraries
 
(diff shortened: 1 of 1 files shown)


FindLibJpegTurbo.cmake
```diff
@@ -18,14 +19,52 @@ find_path(LibJpegTurbo_INCLUDE_DIR
...
+
+# Found the header, read version
+if(_version_header)
+  file(READ "${_version_header}" _header)
+  if(_header)
+   string(REGEX REPLACE ".*#define[\t ]+LIBJPEG_TURBO_VERSION[\t ]+([0-9.]+).*"
+     "\\1" LibJpegTurbo_VERSION "${_header}")
+  endif()
+  unset(_header)
+endif()
```
eile: Does it make sense to assume a given version in the else() path, e.g., 1.2?

rdumusc: I'm not really sure. If the header can't be found, it means jconfig.h is installed in an unexpected location or missing altogether. Those are not good signs and the script will fail if it can't determine the version. Normally users should be able to rely on the LibJpegTurbo_VERSION output. The only benefit of setting a default value would be for users who have a broken or unusual install AND don't care about version checking, but then the only possible default value would be 0 or empty IMO.

PR #545, opened 2017-03-16
Improve packaging of Linux / .deb GUI applications

rdumusc:
Preparation work to release a Deflect package with a DesktopStreamer menu entry for Ubuntu 16.04

### DIFF
--- a/CommonApplication.cmake
+++ b/CommonApplication.cmake
@@ -17,8 +17,10 @@
 # * NAME_OMIT_CHECK_TARGETS do not create cppcheck targets
 # * ARGN for optional add_executable parameters
 # * NAME_DATA files for share/Project/data (in binary and install dir)
-# * NAME_DESKTOP optional .desktop file (Linux GUI applications only)
+# * NAME_DESKTOP optional .desktop file (Linux GUI applications only).
+#   Default: "lowercase(Name).desktop" if the file exists
 # * NAME_ICON optional .icns (Mac OS) or .png (Linux) file (GUI app. only)
+#   Default: "lowercase(Name).{icns|png}" if the file exists
 # * NAME_COPYRIGHT optional copyright notice (Mac OS GUI applications only)
 #
 # Builds Name application, generates doxygen help page and installs it.
@@ -69,8 +71,15 @@ function(common_application Name)
   install(TARGETS ${Name} DESTINATION bin COMPONENT apps)
 
   if(THIS_GUI AND NOT APPLE AND NOT MSVC)
+    if(NOT DEFINED ${NAME}_ICON AND EXISTS ${name}.png)
+      set(${NAME}_ICON ${name}.png)
+    endif()
     if(${NAME}_ICON)
       install(FILES ${${NAME}_ICON} DESTINATION share/pixmaps COMPONENT apps)
+    endif()
+
+    if(NOT DEFINED ${NAME}_DESKTOP AND EXISTS ${name}.desktop)
+      set(${NAME}_DESKTOP ${name}.desktop)
     endif()
     if(${NAME}_DESKTOP)
       install(FILES ${${NAME}_DESKTOP} DESTINATION share/applications
@@ -79,6 +88,9 @@ function(common_application Name)
   endif()
 
   if(THIS_GUI AND APPLE)
+    if(NOT DEFINED ${NAME}_ICON AND EXISTS ${name}.icns)
+      set(${NAME}_ICON ${name}.icns)
+    endif()
     if(${NAME}_ICON)
       set_source_files_properties(${${NAME}_ICON}
         PROPERTIES MACOSX_PACKAGE_LOCATION Resources)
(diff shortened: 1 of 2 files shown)


rdumusc:
one thing I am not sure about is whether or not we want to keep the ABI version in the package name. In this case it feels a bit strange that users have to install "deflect6" and not "deflect".

hernando:
I've seen the ABI as part of the package name in the past and this is in the guidelines for packaging shared libraries http://packaging.ubuntu.com/html/libraries.html.

To be fully correct, the output packages should be libdeflect6 and libdeflect-dev maybe leaving deflect as a meta package pointing to the other two.
Generating different packages for different components is possible, I was doing it long time ago and CMake has support for this but I can't tell how difficult is it to apply that to our software.

rdumusc:
You're right, this is indeed the correct answer. I also remember that CPack can choose which components to install, so excluding the DEV part for example is easy, but creating the multiple packages which depend on each other might be more tricky. All of this is a bit of an overkill for such a simple project.
In this case, I think that releasing a single "deflect" package would be the simplest solution. 

On a side note, I find it odd the the -dev package doesn't contain the SONAME too...

hernando:
Note that SONAME and ABI number are not the same thing. Packages do not contain the SONAME.

Based on my experience I guess that the -dev packages depends on the library package with the largest ABI version, which means that at system level you can have multiple binary versions, but as a developer your not supposed to use anything but the latest. This makes sense because headers go all under /usr/include so it's not possible to avoid name collisions in the general case, but you don't have this problem with library names.
On the other hand, there are libraries that contain the ABI version in the -dev package (e.g. libjpeg8-dev libjpeg9-dev, libjpeg62-dev) but I'm not sure if this is supposed to be sane.

rdumusc:
Sorry, yes I meant the ABI number.
The rest makes sense. Note that on my system, there is a libjpeg-dev meta-package which depends on libjpeg8-dev - but a libjpeg9-dev is also available for users who need to upgrade.
I googled again and couldn't find a solution to generate proper multiple packages using CPack.
Now I'm still not sure what the single package should be named. Options include:
deflect
deflect6
libdeflect6
libdeflect6-dev
Any opinions? Juan you were saying that deflect6 has your preference? For me just "deflect" looks like the most intuitive for apt-get users and still gives the possibility to split the package into components in the future. libdeflect* is probably incorrect at this point since the package contains more than the lib (it also includes the application(s)).

tribal-tec:
Ideally there would be desktopstreamer package which depends on libdeflect/deflect/[version] package. For end-users deflect does not give away that this contains the desktopstreamer tool they want.

CommonApplication.cmake
```diff
@@ -17,7 +17,8 @@
 # * NAME_OMIT_CHECK_TARGETS do not create cppcheck targets
 # * ARGN for optional add_executable parameters
 # * NAME_DATA files for share/Project/data (in binary and install dir)
-# * NAME_ICON optional .icns file (Mac OS GUI applications only)
+# * NAME_DESKTOP optional .desktop file (Linux GUI applications only)
+# * NAME_ICON optional .icns (Mac OS) or .png (Linux) file (GUI app. only)
```
eile: The name should be without extension, so user CMake code is simpler. We can append the extension here and document it.

hernando:
desktopstreamer seems future-proof, users don't care about Deflect and if we ever split the packages in the future the package that matters to users will stay with the same name.

rdumusc:
@tribal-tec That's true but it requires one another separate package with more inter-dependencies. I tired to play with component grouping with CPack to generate multiple packages but I can't modify the package output names (currently deflect6_0.13.0~xenial_amd64-Development.deb and deflect6_0.13.0~xenial_amd64-Runtime.deb) . Looks like it might be possible with CMake >= 3.6, but on Ubuntu 16.04 we only have 3.5... I'm going to install a newer cmake to test that, but since this is ultimately going to run on the Ubuntu VMs it might not be an option.
@hernando good idea, I think that's the best solution for now!

hernando:
Let me search what I did in RTNeuron, because I'm pretty sure I was able to create packages exactly with the names I wanted and the correct dependencies. My only concern is that I may have used some secret magick sauce and if the recipe is not in the repository, then it's certainly lost.

CommonApplication.cmake
rdumusc: Why not, we can do that, but it is also less explicit. We could also make it fully auto-magic and search for a file ${app_name}.extension in the current folder...

rdumusc:
For the record, @hernando pointed out to a solution for creating multiple packages with any CMake version, which is to call the *cpack* command line tool multiple times from a custom target. This could be useful to have *package* targets for subprojects too, insead of only a top level *package* target.

But I'm not going to invest time into such a big change until we need it for more projects. As discussed above I will close this story by making a single package named *desktopstreamer*.

rdumusc:
ping - please review the last two commits only, one of which is a counter-proposal to @eile 's request that NAME_ICON should not take an extension

eile:
+1

PR #547, opened 2017-04-06
common_application(GUI): fix finding icon and .desktop file, cleanups

rdumusc:


CommonApplication.cmake
```diff
@@ -38,62 +39,66 @@ function(common_application Name)
...
 
-  if(THIS_WIN32)
-    if(MSVC)
-      set(OPTIONS WIN32)
+    # _ICON is also used to configure Info.plist on OSX
+    # EXISTS only works with absolute file path
+    if(DEFINED ${NAME}_ICON)
+      set(_ICON ${${NAME}_ICON})
+    elseif(EXISTS ${CMAKE_CURRENT_LIST_DIR}/${_icon_file_candidate})
+      set(_ICON ${_icon_file_candidate})
```
hernando: lower case as the others?

rdumusc: yes, I used lower case and added set(_ICON ${_icon})
before configuring the Info.plist file, it is much clearer that way

PR #548, opened 2017-04-12
CommonCPack: refactoring with improved documentation and features

rdumusc:
I tried to clean up while keeping as much backward compatibility as possible.
See example usage here:
https://github.com/BlueBrain/Tide/compare/master...rdumusc:master
https://github.com/BlueBrain/Deflect/compare/master...rdumusc:master

### DIFF
--- a/Common.cmake
+++ b/Common.cmake
@@ -6,8 +6,8 @@
 #   script is parsed by top level project. See InstallDependencies for details.
 #
 # Output variables
-# - UPPER_PROJECT_NAME - lower-case ${PROJECT_NAME}
-# - LOWER_PROJECT_NAME - upper-case ${PROJECT_NAME}
+# - UPPER_PROJECT_NAME - upper-case ${PROJECT_NAME}
+# - LOWER_PROJECT_NAME - lower-case ${PROJECT_NAME}
 # - TRAVIS - if environment is travis build system
 # - BLUEGENE - if machine is BlueGene
 # - LINUX - if machine is Linux
--- a/CommonCPack.cmake
+++ b/CommonCPack.cmake
@@ -1,225 +1,276 @@
 # Copyright (c) 2013-2017 Stefan.Eilemann@epfl.ch
 #                         Raphael.Dumusc@epfl.ch
+#
 # Info: http://www.itk.org/Wiki/CMake:Component_Install_With_CPack
 #
 # Configures the packaging of the project using CPack.
 #
 # Also includes CommonPackageConfig (legacy, to be removed in the future).
 #
-# Input:
-# * COMMON_PACKAGE_ABI add the ABI version to the package name (default: ON)
+# Input
+# * UPPER_PROJECT_NAME           used as "NAME" for other input below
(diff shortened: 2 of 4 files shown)


Common.cmake
```diff
@@ -6,8 +6,8 @@
 #   script is parsed by top level project. See InstallDependencies for details.
 #
 # Output variables
-# - UPPER_PROJECT_NAME - lower-case ${PROJECT_NAME}
-# - LOWER_PROJECT_NAME - upper-case ${PROJECT_NAME}
+# - UPPER_PROJECT_NAME - upper-case ${PROJECT_NAME}
+# - LOWER_PROJECT_NAME - lower-case ${PROJECT_NAME}
```
eile: OMG

rdumusc:
@eile ping, will you have time to review this today or should I reassign it?

CommonCPack.cmake
```diff
@@ -1,225 +1,276 @@
...
 endif()
 
-include(LSBInfo)
-
 if(CMAKE_SYSTEM_NAME MATCHES "Linux")
-  find_program(RPM_EXE rpmbuild)
-  find_program(DEB_EXE debuild)
-endif()
+  # Legacy?
+  if(EXISTS ${PROJECT_SOURCE_DIR}/CMake/${PROJECT_NAME}.in.spec)
```
tribal-tec: iirc, Equalizer had once a .in.spec file. So probably not needed anymore.

CommonCPack.cmake
```diff
@@ -1,225 +1,276 @@
...
-  set(CPACK_GENERATOR "RPM")
-elseif(DEB_EXE)
-  set(CPACK_GENERATOR "DEB")
-elseif(RPM_EXE)
-  set(CPACK_GENERATOR "RPM")
-else()
-  set(CPACK_GENERATOR "TGZ")
+  # Convert PACKAGE_[REPLACES|CONFLICTS] list to string format without duplicate
+  set(_package_replaces ${${UPPER_PROJECT_NAME}_PACKAGE_REPLACES})
+  list(APPEND _package_replaces ${_old_packages}) # TODO is this really needed?
```
tribal-tec: not sure about this one either?!

CommonCPack.cmake
rdumusc: It still does, along with Collage and Lunchbox. Looking at the file it must be for packaging with native debian tools, not CPack, so here is not the most appropriate place to put the configure() call but I am not sure where else to put it. What do think? Should I move it somewhere else or just remove it?

CommonCPack.cmake
rdumusc: It's hard to say, I don't think regular debian ubuntu packages do that, but they also follow more rules (dev and lib are split, etc). I could try to make some tests to see if dpkg accepts to install libdeflect7 on top of libdeflect6 without this to see if it makes a difference...

rdumusc:
I'm merging as it is to finish the tasks that are waiting on this. We can make another PR later for the 2 points above.

PR #554, opened 2017-06-02
CommonGraph: improve graph output and options, fix recursion

rdumusc:
New ${PROJECT_NAME}_IS_METAPROJECT option in SubProject.cmake to generate graphs without a top-level item for meta-projects that don't call common_find_package(), such as VizStable.


rdumusc:
You can see all the output at /nfs4/bbp.epfl.ch/media/DisplayWall/visualization/Images/software_dependency_graphs

PR #555, opened 2017-06-02
Make generated version file header-only to ease integration w/o need for linking

tribal-tec:


eile:
This breaks the runtime ABI compat check, since the methods are no longer compiled into the lib. Not sure if that's still needed, but then at least ' Runtime check for ABI compatibility.' has to be removed too.

tribal-tec:
Good point. I can keep the cpp then. I just needed the two missing defines actually, I thought I just do  more.

PR #556, opened 2017-06-07
Move -install project chaining to a more common place (fixes latest)

eile:


eile:
The issue is with optional dependencies which are not add_subproject()... Concretely in https://github.com/Eyescale/Equalizer/blob/master/examples/eqPly/CMakeLists.txt#L14, I could also add the dependency ```Equalizer-install ZeroEQ-install ZeroBuf-install``` here, or make Zero* subprojects in Equalizer. Without at least one of the three options, the smoke test fails since libZeroEQ is missing.

rdumusc:
Ah of course! There can't be an install dependency on a library for which you don't add the sources...

I found more other issues with the uxmal change:
common_find_package(ZeroEQ)
common_find_package(ZeroBuf)
are hidden in examples/CMakeLists.txt instead of being in the top-level CMakeLists.txt, which happens after common_find_package_post().

And I wonder why uxmal has a private dependency on ZeroEq? and a non-PUBLIC one on ZeroBuf?
set(UXMAL_LINK_LIBRARIES ZeroBuf ZeroEQ)

And what are CULLBUG_HEADERS? A grep on viz/stable gave me no results:
set(UXMAL_PUBLIC_HEADERS ${CULLBUG_HEADERS})

eile:
> install dependency on a library for which you don't add the sources

In latest we can! ;)

Let's see tomorrow. I wanted to avoid bloating Eq dependencies, maybe it's time to move uxmal to a separate project.

tribal-tec:
+1

CommonInstallProject.cmake
```diff
@@ -0,0 +1,27 @@
...
+
+if(NOT TARGET ${PROJECT_NAME}-install)
+  add_custom_target(${PROJECT_NAME}-install
+    ${CMAKE_COMMAND} -P ${PROJECT_BINARY_DIR}/cmake_install.cmake
+    DEPENDS ${PROJECT_NAME}-all)
+  set_target_properties(${PROJECT_NAME}-install PROPERTIES
+    EXCLUDE_FROM_DEFAULT_BUILD ON)
+endif()
+
+# redo dependency chaining each time: a new common_find_package might have been called
```
rdumusc: the first inclusion of this file will be in Common.cmake, before any common_find_package() call. So this will only work if users include DoxygenRule.cmake at the end... A solution I can think of is to make this a function to be called from common_find_package_post() instead.

CommonInstallProject.cmake
```diff
@@ -0,0 +1,27 @@
+# Copyright (c) 2017 Stefan.Eilemann@epfl.ch
+#
+# Generates ${PROJECT_NAME}-install rule. reentrant, to be called by every
+# module depending on ${PROJECT_NAME}-install.
+#
+# This rule is for targets which depend on the installed artefacts, e.g.,
+# doxygen and smoke tests. CMake does not provide this, so this is a workaround.
+# install naturally depends on all, which depends on all common_library and
+# common_application targets. Furthermore, ${PROJECT_NAME}-install depends on
+# all subprojects which have a install rule, so that all necessary artefacts in
```
rdumusc: an install rule

CommonInstallProject.cmake
```diff
@@ -0,0 +1,27 @@
...
+# a superproject build are installed.
```
rdumusc: document ${PROJECT_NAME}_FIND_PACKAGES_FOUND as input variable filled by common_find_package()

DoxygenRule.cmake
```diff
@@ -64,6 +64,7 @@ if(NOT TARGET ${PROJECT_NAME}-all)
 endif()
 
 include(CommonDate)
+include(CommonInstallProject)
```
rdumusc: \n

CommonInstallProject.cmake
eile: done

CommonInstallProject.cmake
eile: done

rdumusc:
On second thought, I still think that the real problem is Equalizer trying to hide its subproject dependencies. This workaround is for a use case we should not support, because the hidden dependencies bypass the CI checks but may fail later in the super-project case. CommonFindPackage does not have anything to do with -install targets, which are project-to-project relationships for the Subprojects use case.

CommonInstallProject.cmake
eile: ? See doc: ```to be called by every module depending on ${PROJECT_NAME}-install.```

DoxygenRule.cmake
eile: There was no newline before, but if it makes you happy.

eile:
No, the real problem is that CMake does not provide a Target-install target which installs the target fully. I'm happy to explain the details in person.

eile:
ping

CommonInstallProject.cmake
```diff
@@ -0,0 +1,31 @@
...
+if(NOT TARGET ${PROJECT_NAME}-install)
+  add_custom_target(${PROJECT_NAME}-install
+    ${CMAKE_COMMAND} -P ${PROJECT_BINARY_DIR}/cmake_install.cmake
+    DEPENDS ${PROJECT_NAME}-all)
+  set_target_properties(${PROJECT_NAME}-install PROPERTIES
+    EXCLUDE_FROM_DEFAULT_BUILD ON)
+endif()
+
+function(COMMON_ADD_INSTALL_DEPENDENCIES)
+  # redo dependency chaining each time: a new common_find_package might have been called
```
rdumusc: can be removed now

CommonInstallProject.cmake
```diff
@@ -0,0 +1,31 @@
...
+  string(REPLACE " " ";" __deps "${ARGN}") # string-to-list
```
rdumusc: No longer needed I think, CMake expends the list when calling the function

CommonInstallProject.cmake
eile: done

CommonInstallProject.cmake
eile: It won't. I tested.

PR #558, opened 2017-07-20
Fix subtle clang-format foo

eile:


hernando:
I hate these commit message that say nothing. This commit seems to be bogus I can't remember what was the reason for it, grrr

eile:
It's not bogus. If there is no .clang-format, it would fall back to the default style and fuck things up. Even though that should never happen, it did sometimes.

hernando:
Haha, we were wondering yesterday whether you would reply to the comment :P

@tribal-tec mentioned that the command line option added in this commit is causing clang-format to not do anything at all in a recent version. I guess he can provide more info.
I remember seeing bizarre behavior in your system when you submitted this patch, but I can't rememeber what was it exactly. If there's no .clang-format file, the script exits earlier, how is it possible to get clang-format run under such a condition?

tribal-tec:
from --help:
```Use -fallback-style=none to skip formatting.```

But as @hernando  said, it should never reach that point if there's no .clang-format file.

**Comment by eile** [comment]
@hernando : Now you know! ;)
@tribal-tec : Also from --help:
```
fallback in case clang-format is invoked with -style=file, but can not find the .clang-format file to use.
```

... which we do!? In any case, it does not matter for me (at least in the foreseeable future).

PR #560, opened 2017-07-24
Simplify GitExternals

eile:


eile:
@hernando : Please remove from Pydoxine and SDK.

rdumusc:
Some of the above are likely false-positives because ninja-rebase has undefined behaviour for me ATM.

PR #565, opened 2017-09-12
Fix bitrot in rpm spec license config

eile:


eile:
Needed for https://travis-ci.org/Eyescale/Lunchbox/jobs/274536043

PR #576, opened 2018-01-13
Use cuda_add_library and cuda_add_executable if cuda is found

jjcasmar:


hernando:
It looks this will work for your case but could have undesired side effects in the general case. In particular, in complex projects that have multiple targets, you don't want to use cuda_add_foo for those ones that don't have a direct dependency on CUDA. Can you confirm what happens if you add common_find_package(CUDA) to a project that actually doesn't use it?

jjcasmar:
You are absolutely right. For a particular target that doesn't need CUDA, cuda_add_foo would still be called, which may be undesirable. Maybe we can activate cuda calls by setting a flag for the target set(TARGET_USE_CUDA). 

I have test this with a project which has some targets needing CUDA and some targets that don't and it works fine anyway, but, of course, its not perfect.

PR #593, opened 2018-12-18
Fix clang-format & CommonFindPackage

tribal-tec:


CommonFindPackage.cmake
```diff
@@ -47,6 +47,11 @@ option(COMMON_FIND_PACKAGE_QUIET "Use QUIET for common_find_package command" ON)
 macro(common_find_package Package_Name)
   string(TOUPPER ${Package_Name} PACKAGE_NAME)
 
+  # ensure outputs vars exists
+  set(COMMON_FIND_PACKAGE_DEFINES "")
```
hernando: This one doesn't seem correct. You're overriding any previous value it may have. I would check `if (COMMON_FIND_PACKAGE_DEFINES)` in common_find_package_disable instead.

tribal-tec: It's first used here. I added the missing `if()` for all the vars.

PR #594, opened 2019-02-25
Use CMake 3.13 backwards compatible syntax for policy handling

hernando:
More details here https://cliutils.gitlab.io/modern-cmake/chapters/basics.html

### DIFF
--- a/Common.cmake
+++ b/Common.cmake
@@ -19,16 +19,17 @@
 # - A <project>-all target to build only the given (sub)project
 # - A <project>-install target to build and install the given (sub)project
 
-cmake_minimum_required(VERSION 3.1 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.1...3.13 FATAL_ERROR)
 
 if(CMAKE_INSTALL_PREFIX STREQUAL PROJECT_BINARY_DIR)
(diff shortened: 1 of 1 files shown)


Common.cmake
```diff
@@ -19,16 +19,17 @@
 # - A <project>-all target to build only the given (sub)project
 # - A <project>-install target to build and install the given (sub)project
 
-cmake_minimum_required(VERSION 3.1 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.1...3.13 FATAL_ERROR)
```
tribal-tec: why the upper limit? 3.14 is already out, so for every new release this has to be touched?

hernando: I didn't know 3.14 had already been released. I have tested this with 3.13 so I cannot claim that 3.14 is supported. When this script is executed with a CMake version higher than 3.13 it won't fail, in this case CMake will leave all policies unknown to 3.13 unset  (more details here: https://cmake.org/cmake/help/v3.13/command/cmake_minimum_required.html#policy-settings) If we set max too high, we are implicitly accepting the new behaviour of any future policy, which may lead to unnoticed bugs as we will never see the warning about these policies being unset.

hernando: ping

PR #595, opened 2019-06-11
Fix for Ubuntu bionic lsb_codename

ppodhajski:
With Ubuntu 18.04 apt behaves differently wrt to codename appended (with ~)  to package version.

In the example desktopstreamer 1.0.2 is  not the same as desktopstreamser 1.0.2~bionic

`The following packages have unmet dependencies:
 tide : Depends: desktopstreamer (>= 1.0.2) but 1.0.2~bionic is to be installed`


CommonCPack.cmake
```diff
@@ -202,12 +202,11 @@ if(CPACK_GENERATOR STREQUAL "DEB")
   # Note: the ~codename is not part of any standard and could be omitted.
   if(NOT CPACK_DEBIAN_PACKAGE_VERSION)
     include(LSBInfo)
-    set(CPACK_DEBIAN_PACKAGE_VERSION "${CPACK_PACKAGE_VERSION}~${LSB_CODENAME}")
-  endif()
-  if(NOT CPACK_PACKAGE_FILE_NAME)
```
tribal-tec: why removed this if as well?

ppodhajski: let's say it was my editor ;-)

